PR #25, opened 2019-02-07
Add RabbitMQ template

GregaVrbancic:
I have added the template for RabbitMQ and tested it as instructed in README.

### DIFF
--- a/public/v1/apps/rabbitmq.json
+++ b/public/v1/apps/rabbitmq.json
@@ -19,7 +19,7 @@
         }
     },
     "instructions": {
-        "start": "RabbitMQ is open source message broker software (sometimes called message-oriented middleware) that implements the Advanced Message Queuing Protocol (AMQP). The RabbitMQ server is written in the Erlang programming language and is built on the Open Telecom Platform framework for clustering and failover. Client libraries to interface with the broker are available for all major programming languages.\n\n Enter your Postgres Configuration parameters and click on next. It will take about a minute for the process to finish.",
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

githubsaturn:
Published.

githubsaturn:
@GregaVrbancic 

I have no experience with RabbitMQ. But from their docs, it looks like `hostname` is actually pretty important - something regarding Node Name and how it stores data...

CapRover one-click-apps don't support explicit hostname. But the good news is that by the looks of it, `--hostname` is not the only way that RabbitMQ can accept the Node Name. We can also set an environmental variable, `RABBITMQ_NODENAME`, with proper value. We can default to `rabbit@localhost`.

What do you think? Is this `hostname` and Node Name actually important? If so, I think one way or another we need to fix this.

GregaVrbancic:
@githubsaturn thank you for pointing out the issue. I have missed that one. Sure, defaulting to the `rabbit@localhost` using the `RABBITMQ_NODENAME` environmental variable seems right. 

I have already tested the fixed template with the mentioned solution implemented on my CapRover instance and it seems to be working correctly, giving the following log output:

```
2019-02-07T07:53:33.011917705Z 2019-02-07 07:53:33.011 [info] <0.208.0> 
2019-02-07T07:53:33.011962448Z node : rabbit@localhost
2019-02-07T07:53:33.011978481Z home dir : /var/lib/rabbitmq
2019-02-07T07:53:33.011993622Z config file(s) : /etc/rabbitmq/rabbitmq.conf
2019-02-07T07:53:33.012008463Z cookie hash : dMCFt1nm//JVYeGyeucaFw==
2019-02-07T07:53:33.012023452Z log(s) : <stdout>
2019-02-07T07:53:33.012038631Z database dir : /var/lib/rabbitmq/mnesia/rabbit@localhost
```

We can see that the database dir has the `rabbit@localhost` appended to the end of the database dir, as is presented in the RabbitMQ documentation.

If you agree, I will prepare a new pull request addressing this issue.

PR #34, opened 2019-03-16
Upgrade Thumbor

gingerlime:
* APSL/thumbor is not actively maintained and doesn't have the latest versions
* [MinimaCompact](https://github.com/MinimalCompact/thumbor) is actively maintained and works closely with the thumbor core team
* This change switches to use minimalcompact and also upgrades the default version to the latest
* Note: not sure if it's worth hard-coding an explicit version `6.7.0` or use the `latest` tag?

### DIFF
--- a/public/v1/apps/thumbor.json
+++ b/public/v1/apps/thumbor.json
@@ -1,11 +1,11 @@
 {
     "captainVersion": "1",
-    "documentation": "Taken from https://hub.docker.com/r/apsl/thumbor/tags/",
+    "documentation": "Taken from https://hub.docker.com/r/minimalcompact/thumbor/tags/",
     "dockerCompose": {
         "version": "3.3",
         "services": {
             "$$cap_appname-db": {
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

> Note: not sure if it's worth hard-coding an explicit version 6.7.0 or use the latest tag?

we always use explicit version to make sure the result is stable and predicable.

gingerlime:
Thanks. Sounds like a good approach. I'll try to remember to submit a PR when a new version is released.

An unrelated question, but I was wondering about it... Is it possible to actually copy&paste the one-click-app json, the same way you can paste a captain definition file? I just thought it could be useful for testing out or slightly modifying one-click apps.

githubsaturn:
For sure, see here: https://github.com/caprover/one-click-apps#test-your-one-click-apps

gingerlime:
Cool. Thanks! Sorry I missed it.

PR #36, opened 2019-03-17
Create hasura.json

comontes:
This is my first time working with a docker-compose file, but this seems to work fine. Original file taken from https://github.com/hasura/graphql-engine/tree/master/install-manifests/docker-compose-postgis this version has Postgis included


githubsaturn:
Thanks!

PR #39, opened 2019-03-17
Fix problem with port number on Ghost

felipe-augusto:
For some reason, the port was defaulting to 2368 and it was constructing the application URLs wrong.

Instead of `http://cap-app.myroot.example.com/`,
it was generating the url `http://cap-app.myroot.example.com:2368/`

Also, I have two suggestions, I think this is not the right place to do it, but here we go:

1. I think it's a good ideia to change the select component from the one-click-app UI, to allow searching, because the number of apps is increasing.  https://ant.design/components/select/

2. When building this template I originally used the official ghost image instead of the image from bitnami. But I had a problem using env variables to configure SMTP options. In the original [config](https://docs.ghost.org/concepts/config/#mail) docs it goes like this:

```javascript
"mail": {
    "transport": "SMTP",
    "options": {
        "service": "Mailgun",
        "auth": {
            "user": "postmaster@example.mailgun.org",
            "pass": "1234567890"
        }
    }
}
```

But I didn't know how to configure this using `env` variables. What do you think is the default behavior when the config looks like that?

### DIFF
--- a/public/v1/apps/ghost.json
+++ b/public/v1/apps/ghost.json
@@ -1,0 +1,150 @@
+{
+   "captainVersion":"1",
+   "documentation":"Taken from https://docs.ghost.org/",
+   "dockerCompose":{
+      "services":{
+         "$$cap_appname":{
+            "containerHttpPort":"2368",
+            "environment":{
+               "GHOST_DATABASE_NAME":"ghost",
+               "GHOST_DATABASE_PASSWORD":"$$cap_db_ghost_password",
+               "GHOST_DATABASE_USER":"ghost",
+               "GHOST_EMAIL":"$$cap_ghost_email",
+               "GHOST_HOST":"$$cap_ghost_host",
+               "GHOST_PASSWORD":"$$cap_ghost_password",
+               "GHOST_PROTOCOL":"$$cap_ghost_protocol",
+               "MARIADB_HOST":"srv-captain--$$cap_appname-db",
+               "MARIADB_PORT_NUMBER":"3306",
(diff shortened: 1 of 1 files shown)


githubsaturn:
> I think it's a good ideia to change the select component from the one-click-app UI, to allow searching, because the number of apps is increasing. https://ant.design/components/select/

Agreed. Let me know if you want to take that task on. Otherwise, please create an issue on the main repo and I, or someone else, will take it on.

> When building this template I originally used the official ghost image instead of the image from bitnami. But I had a problem using env variables to configure SMTP options. In the original config docs it goes like this:

Ghost has an unconventional scheme for nested config values defined as env, see this:
https://docs.ghost.org/concepts/config/#running-ghost-with-config-env-variables

Basically to set the config that you mentioned above, you have to set env vars like this:
mail__transport: SMTP
main__options__service: Mailgun

githubsaturn:
Port change is good. But you also needed to remove `containerHttpPort`. I'll do that myself after merging :+1:

felipe-augusto:
@githubsaturn I think that you need to keep the `containerHttpPort`. I tested it here and it's broken.
This [comment](https://github.com/bitnami/bitnami-docker-ghost/issues/56#issuecomment-317385441) here explains:

> The ghostPort parameter specifies the port number that should be used while constructing the application URLs. It does not change the port on which the ghost app server listens on.

githubsaturn:
Ah, my bad. Fixed.

githubsaturn:
Just pushed the fix. Wait a minute or two before retrying.

PR #44, opened 2019-03-29
adding drone.io build automation

tlwt:
see drone.io for more info

### DIFF
--- a/public/v1/apps/drone.json
+++ b/public/v1/apps/drone.json
@@ -39,7 +39,7 @@
     },
     "instructions": {
         "start": "Drone is a self-service Continuous Delivery platform for busy development teams.\n\n Enter your Drone for Gitea configuration parameters and click on next. The process will take about a minute to finish.",
-        "end": "Drone is deployed and available as $$cap_appname-server.\n\n  IMPORTANT: It will take up to 2 minutes for Gitea to be ready. Before that, you might see 502 error page.\n"
+        "end": "Drone is deployed and available as $$cap_appname-server.\n\n  IMPORTANT: It will take up to 2 minutes for Drone for Gitea to be ready. Before that, you might see 502 error page.\n"
     },
     "variables": [
         {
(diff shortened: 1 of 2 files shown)


githubsaturn:
Thanks for your first contribution!

public/v1/apps/drone.json
```diff
@@ -0,0 +1,102 @@
+{
+    "captainVersion": "1",
+    "documentation": "Taken from https://hub.docker.com/r/drone/drone/",
```
githubsaturn: This is just a link to their Docker Hub page, it doesn't include any docs. Usually we use this "documentation" field to point to the URL where the original Docker Compose file was taken. This will help us to address potential issues in the future as we can reach out to the original provider and ask for their advice.

If you manually created this Docker Compose file, that's fine too. Just post a link to the docs where this composition is explained (agent and server).

Let me know what you think.

public/v1/apps/drone.json
```diff
@@ -0,0 +1,102 @@
...
+                    "DRONE_SERVER_PROTO": "$$cap_drone_server_proto",
+                    "DRONE_SERVER_HOST": "$$cap_drone_server_host",
+                    "DRONE_TLS_AUTOCERT": "$$cap_drone_tls_autocert",
+                    "DRONE_SECRET": "$$cap_drone_secret"
+                }
+            }
+        }
+    },
+    "instructions": {
+        "start": "Drone is a self-service Continuous Delivery platform for busy development teams.\n\n Enter your Drone for Gitea configuration parameters and click on next. The process will take about a minute to finish.",
```
githubsaturn: It seems like Drone can work with many git clients (including Gitea). But this drone config is geared only towards Gitea.

I might be wrong. But if my assumption above is correct, let's:
- rename this file to `drone-gitea` instead of `drone` 
- Add clear instructions in `instructions.start` saying that you need to have an already running Gitea server. This does not automatically spin up a Gitea instance for you.

public/v1/apps/drone.json
tlwt: agreed

public/v1/apps/drone.json
tlwt: agreed

tlwt:
Pushed to ```tlwt/one-click-apps```.

githubsaturn:
Thanks!

PR #45, opened 2019-04-03
Prisma with MySQL

chriz001:

### DIFF
--- a/public/v1/apps/prisma.json
+++ b/public/v1/apps/prisma.json
@@ -26,7 +26,7 @@
                     "$$cap_appname-db"
                 ],
                 "environment": {
-                    "PRISMA_CONFIG": "port: 4466\ndatabases:\n  default:\n    connector: mysql\n    host: srv-captain--$$cap_appname-db\n    port: 3306\n    user: root\n    password: $$cap_db_pass\n    migrations: true\n"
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks for your first contribution!

public/v1/apps/prisma.json
```diff
@@ -0,0 +1,69 @@
...
+        },
+        {
+            "id": "$$cap_db_pass",
+            "label": "Database password",
+            "description": "",
+            "validRegex": "/.{1,}/"
+        },
+        {
+            "description": "To ensure only entitled users are able to perform actions through the Management API",
+            "id": "$$cap_management_secret",
```
githubsaturn: This value is not being set. I think it needs to be added to `PRISMA_CONFIG` env var.

public/v1/apps/prisma.json
```diff
@@ -0,0 +1,69 @@
...
+            "label": "Prisma Management API secret",
+            "validRegex": "/^(?=.*\\d).{10,}$/"
```
githubsaturn: Add a bit to description explaining this regex to avoid user confusion. Something like:
```
Minimum 10 characters, at least one letter and one number.
```

public/v1/apps/prisma.json
chriz001: Good pickup, thanks.

public/v1/apps/prisma.json
chriz001: Done

githubsaturn:
Awesome, thanks!

PR #46, opened 2019-04-06
feat(*): add template for verdaccio

aldaircruz:


githubsaturn:
Thanks for your contributions!

PR #50, opened 2019-04-08
add elasticsearch (single and cluster)

fraank:
receipt to add elasticsearch for single and multi node usage.

i'm aware that kibana and logstash (which are mostly used together as ELK-stack) are missing. i would add them as separate receipes as they're not necessarily needed for every node.

### DIFF
--- a/public/v1/apps/elasticsearch.json
+++ b/public/v1/apps/elasticsearch.json
@@ -67,7 +67,7 @@
             "id": "$$cap_container_port",
             "label": "Container TCP Port",
             "defaultValue": "9200",
-            "description": "Port of of Elasticsearch to be exposed by the container.",
+            "description": "Internal port for Elasticsearch the container listens to.",
             "validRegex": "/^([0-9])+$/"
         }
     ]
--- a/public/v1/apps/elasticsearch.json
+++ b/public/v1/apps/elasticsearch.json
@@ -1,0 +1,83 @@
+{
+    "captainVersion": 1,
+    "documentation": "Elasticsearch (Single and Cluster). Taken from docker.elastic.co/elasticsearch/elasticsearch",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname": {
+                "image" : "docker.elastic.co/elasticsearch/elasticsearch:$$cap_elasticsearch_version",
+                "notExposeAsWebApp": "true",
+                "volumes": [
+                    "$$cap_appname-elasticsearch-data:/usr/share/elasticsearch/data"
+                ],
+                "restart": "always",
+                "environment": {
+                    "ES_JAVA_OPTS": "-Xms512m -Xmx512m",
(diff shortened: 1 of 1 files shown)


public/v1/apps/elasticsearch.json
```diff
@@ -0,0 +1,83 @@
...
+                    "ES_JAVA_OPTS": "-Xms512m -Xmx512m",
+                    "cluster.name": "$$cap_elasticsearch_cluster_name",
+                    "http.port": "$$cap_container_port",
+                    "node.master": "$$cap_elasticsearch_node_master",
+                    "node.data": "$$cap_elasticsearch_node_data",
+                    "discovery.zen.minimum_master_nodes": "$$cap_elasticsearch_minimum_master_nodes",
+                    "discovery.zen.ping.unicast.hosts": "$$cap_elasticsearch_unicast_hosts"
+                },
+                "ports": [
+                    "$$cap_host_port:$$cap_container_port"
```
githubsaturn: You shouldn't need to port forward. As far as I know, port 9200 of Elastic Search is just an HTTP port. In that case:
- Remove `notExposeAsWebApp`
- Remove `cap_host_port`
- Set  `containerHttpPort` to `$$cap_container_port`

public/v1/apps/elasticsearch.json
```diff
@@ -0,0 +1,83 @@
...
+                ]
+            }
+        }
+    },
+    "instructions": {
+        "start": "Elasticsearch search engine. Be sure to extend the virtual memory on all your hosts! (increase \"echo 'vm.max_map_count=262144' >> /etc/sysctl.conf\" and restart \"sysctl -p\")",
+        "end": "Elasticsearch is deployed as $$cap_appname. Goto %%cap_host_port/_cat/health to check the health of your cluster."
```
githubsaturn: - Add a note to `end` saying that the deploy is finished. Elastic Search is starting up. You might see 502 errors for the next 2 minutes.
- Replace `%%cap_host_port` with `<YOUR-APP-URL>`

public/v1/apps/elasticsearch.json
```diff
@@ -0,0 +1,83 @@
+{
+    "captainVersion": 1,
+    "documentation": "Elasticsearch (Single and Cluster). Taken from docker.elastic.co/elasticsearch/elasticsearch",
```
githubsaturn: This is a documentation field. I think your URL is just a link to the Docker image. Can you please replace it with the documentation for Elastic Search? This is for future maintenance... Specifically, we want to document where and how the following items are set:
- Environmental variables such as `discovery.zen.ping.unicast.hosts` , `ES_JAVA_OPTS`,  `cluster.name` and etc
- Volumes path such as `/usr/share/elasticsearch/data`

public/v1/apps/elasticsearch.json
fraank: makes sense.

PR #51, opened 2019-04-10
feat(*): add template for sourcegraph

aldaircruz:


public/v1/apps/sourcegraph.json
```diff
@@ -0,0 +1,42 @@
...
+                "containerHttpPort":"7080",
+                "image": "sourcegraph/server:$$cap_container_version",
+                "volumes": [
+                    "$$cap_appname-config:/etc/sourcegraph",
+                    "$$cap_appname-data:/var/opt/sourcegraph"
+                ],
+                "restart": "always",
+                "environment": {},
+                "ports": [
+                    "$$cap_admin_console_port:2633"
```
githubsaturn: I don't think we need to expose this port. User can manually expose the admin console port anytime they want. Exposing port "may" be a security threat.

githubsaturn:
Thanks for your contribution. Just one minor comment on the port forwarding.

public/v1/apps/sourcegraph.json
aldaircruz: I understand your point, however the administration console requires a basic authentication using a username and a very strong and read-only password generated by the application for this use case.

> This password is automatically generated for you to ensure it is very long and secure, even if the management console is exposed to the public internet.

More details here:
https://docs.sourcegraph.com/admin/management_console#what-is-the-password-to-my-management-console

Can I remove it anyway?

githubsaturn: Sounds good. I have no experience with SourchGraph. My vote is to remove this port if:
- it's not necessary for the end users to use this port.
- this port is only being used in special cases and users can choose to expose the port if they want

Another security concern with the port forwarding on an HTTP port is that users won't be using HTTPS if it's a custom port. This means their password will be transported via a cleartext HTTP request. This is really dangerous. 

The ideal solution would be to deploy two containers. The first container is the SourceGraph. The second container would be a dumb container which just acts as an HTTP forwarder and forward all HTTP requests to SourceGraph container port `2633`. Creating such a container with custom nginx config is easy. But we want a container which simply gets its configuration from env vars, rather than config files. I am not sure if such container exists, but creating it is not too difficult.

If you want, we can merge this PR as is and we can add this second image later. Let me know your thoughts.

aldaircruz: I agree with your points regarding removing the port, because at least for me, I used the console only to disallow new users from signing up.

I'm also not very experienced with Sourchgraph, I've been just exploring and testing it a bit.

I didn’t noticed before, but the management console is only available [using https](https://docs.sourcegraph.com/admin/management_console#i-am-getting-the-server-sent-an-invalid-response-errors-from-my-browser-why) by using a self signed certificate which they provide. 

> The self-signed TLS certificate in use ensures that your interaction with the management console cannot be sniffed via MITM attacks. If desired, you can configure the management console to use your own TLS certificates (but you do not need to).

[How to change the TLS certificate](https://docs.sourcegraph.com/admin/management_console#how-can-i-use-my-own-tls-certificates-with-the-management-console)

It’s possible, but not recommended, to [disable the https](https://docs.sourcegraph.com/admin/management_console#can-i-disable-https-on-the-management-console) on the console (using the console) .

I like this second container idea, but does is it still necessary since the console is only available (by default) using https?

githubsaturn: Interesting thing about self-sign cert... It basically means that we cannot use the second container idea as it acts as an HTTPS terminal point. For simplicity, let's remove this port and add some notes to `instructions.end` to let the user know that they can expose the port and use the management console. 


PS
The second container is still possible, but it requires disabling HTTPS on management and letting nginx manage HTTPS for you. It's not unsafe at all because the unencrypted communication only happens within the server and the unencrypted data never leaves the server. However, this is a more complicated setup and for now we can just use a single container and allow the user to open the management port if they want to do so later on.

aldaircruz: I removed the port and added some notes to `instructions.end` regarding the management console.

githubsaturn:
Awesome! I'll create that HTTP forwarder image asap.

githubsaturn:
PS: I just created the one-click reverse proxy: https://github.com/caprover/one-click-apps/blob/master/public/v1/apps/nginx-reverse-proxy.json

PR #56, opened 2019-04-14
feat(*): add template for gitlab-ce

aldaircruz:

### DIFF
--- a/public/v1/apps/gitlab-ce.json
+++ b/public/v1/apps/gitlab-ce.json
@@ -1,0 +1,39 @@
+{
+    "captainVersion": "1",
+    "documentation": "Taken from https://docs.gitlab.com/omnibus/docker/",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname": {
+                "image": "gitlab/gitlab-ce:$$cap_gitlab_version",
+                "volumes": [
+                    "$$cap_appname-config: /etc/gitlab",
+                    "$$cap_appname-logs: /var/log/gitlab",
+                    "$$cap_appname-data: /var/opt/gitlab"
+                ],
+                "restart": "always",
+                "containerHttpPort": "80",
+                "environment": {
+                    "GITLAB_OMNIBUS_CONFIG": "$$cap_gitlab_omnibus_config"
+                }
+            }
+        }
+    },
+    "instructions": {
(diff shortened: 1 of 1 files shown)


public/v1/apps/gitlab-ce.json
```diff
@@ -0,0 +1,43 @@
...
+                    "$$cap_appname-logs: /var/log/gitlab",
+                    "$$cap_appname-data: /var/opt/gitlab"
+                ],
+                "restart": "always",
+                "containerHttpPort": "80",
+                "environment": {
+                    "GITLAB_OMNIBUS_CONFIG": "$$cap_gitlab_omnibus_config"
+                },
+                "ports": [
+                    "22:22"
```
githubsaturn: The host's SSH is also at port 22. Doing this will can cause the host's SSH to malfunction.

public/v1/apps/gitlab-ce.json
```diff
@@ -0,0 +1,43 @@
+{
+    "captainVersion": "1",
+    "documentation": "Taken from https://hub.docker.com/r/gitlab/gitlab-ce/",
```
githubsaturn: I think this link has better documentation: https://docs.gitlab.com/omnibus/docker/

aldaircruz: Hi, I'm using a [regular expression](https://www.regextester.com/?fam=108965) to only allow a valid TCP port except 22 (already used by caprover) but the result is always not valid. It seems like the world boundary is being removed (at least from the error) by caprover.

Any idea why this happens?

Image:

imagem

githubsaturn: Two things:
1 - The regex you have must be stringified version of the Regex. `\b` must be converted to `\\b`. Run this JS snippet to get the correct formatting for JSON:
```
JSON.stringify(
{
     validRegex : /(\b([1-9]|1[0-9]|2[01])\b)|(\b(2[3-9]|[3-9][0-9]|[1-8][0-9]{2}|9[0-8][0-9]|99[0-9]|[1-8][0-9]{3}|9[0-8][0-9]{2}|99[0-8][0-9]|999[0-9]|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])\b)/.toString()
})
```

2- Similar to the previous PR, let's remove the PORT mapping from gitlab configuration if it's not 100% necessary. Users can set their own port mapping after the app is deployed. Gitlab doesn't rely on this port mapping to work. I suggest we add some notes that users can enable this port mapping after the installation is done - only if they need to. Reason for this suggestion is:
- Port mapping carries security risk as it may bypass the firewall (on some OS). This is specially dangerous if the PORT is not secured.
- In order for SSH to work properly users have to configure their keys anyways. Therefore, just by assigning port mapping, we can't say "SSH" is working as expected...
- Other than port 22, there are other ports used by CapRover or Docker, they have to be illegal to use. See this: https://caprover.com/docs/firewall.html

For these reasons, I suggest we remove the "ports" section and add the explanation to instructions.end.

aldaircruz: Hi, thanks! I just committed the changes!

githubsaturn:
Thanks

PR #60, opened 2019-04-21
feat(*): add template for sentry

aldaircruz:
Hi @githubsaturn,   
Do you have any suggestion?

### DIFF
--- a/public/v1/apps/sentry.json
+++ b/public/v1/apps/sentry.json
@@ -1,0 +1,214 @@
+{
+    "captainVersion": "1",
+    "documentation": "Taken from https://hub.docker.com/_/sentry/ and https://docs.sentry.io/server/",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname-postgres": {
+                "image": "postgres:$$cap_postgres_version",
+                "notExposeAsWebApp": "true",
+                "volumes": [
+                    "$$cap_appname-postgres-data:/var/lib/postgresql/data"
+                ],
+                "restart": "always",
+                "environment": {
+                    "POSTGRES_USER": "sentry",
+                    "POSTGRES_PASSWORD": "$$cap_pg_pass",
+                    "POSTGRES_DB": "sentry"
+                }
+            },
+            "$$cap_appname-redis": {
+                "dockerfileLines": [
+                    "FROM redis:$$cap_redis_version",
+                    "CMD exec redis-server"
+                ],
+                "notExposeAsWebApp": "true",
+                "volumes": [
+                    "$$cap_appname-redis-data:/data"
+                ],
+                "restart": "always"
+            },
+            "$$cap_appname-cron": {
+                "depends_on": [
+                    "$$cap_appname"
+                ],
+                "dockerfileLines": [
+                    "FROM sentry:$$cap_sentry_version",
+                    "RUN groupadd -r $$cap_appname-cron && useradd -r -g $$cap_appname-cron $$cap_appname-cron",
+                    "USER $$cap_appname-cron",
+                    "CMD sentry run cron"
+                ],
+                "notExposeAsWebApp": "true",
+                "restart": "always",
+                "environment": {
+                    "SENTRY_SECRET_KEY": "$$cap_sentry_secret",
+                    "SENTRY_REDIS_HOST": "srv-captain--$$cap_appname-redis"
+                }
+            },
+            "$$cap_appname-worker": {
+                "depends_on": [
+                    "$$cap_appname-cron"
+                ],
+                "dockerfileLines": [
+                    "FROM sentry:$$cap_sentry_version",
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!
I have limited internet access in next 10 days. But I'll review this as soon as possible.

githubsaturn:
@aldaircruz 

This looks exciting!

From the initial look, it looks like a relatively heavy stack... You mentioned it takes around 10 minutes to complete the installation process...

What's the minimum RAM/CPU requirements?

aldaircruz:
@githubsaturn 
> 
> This looks exciting!
> 
> From the initial look, it looks like a relatively heavy stack... You mentioned it takes around 10 minutes to complete the installation process...
> 
> What are minimum RAM/CPU requirements?

The initial configuration took those minutes due to the  initial [database migration](https://docs.sentry.io/server/installation/docker/#running-migrations).

The documentation doesn't specify [minimum hardware requirements](https://docs.sentry.io/server/installation/#hardware) values.

The project's [README.md](https://github.com/getsentry/onpremise#minimum-hardware-requirements) mentions the need of having at least [**3GB of ram**](https://github.com/getsentry/onpremise/pull/165). Not filling this requirement may cause some errors during the initial configuration when running the database migrations.

I will add this information to instructions.start.

MeXaaR:
You will end up with an error when integrating some stuff (i.e GitLab) if this is not filled up with the domain of the application
`"RUN echo 'system.url-prefix: \"\"' >> /etc/sentry/config.yml",`

I changed the template with the domain of the app and it works fine after that. So maybe some changes to do on that

githubsaturn:
Is this what you're suggesting? That this line
```
"RUN echo 'system.url-prefix: \"\"' >> /etc/sentry/config.yml",
```
should be changed to something like:
```
"RUN echo 'system.url-prefix: \"my-app-name.domain.com\"' >> /etc/sentry/config.yml",
```


If this is what you are proposing, unfortunately it's not possible at the moment that we dynamically include the domain name in the template. However, it's definitely something that we can add to the template system. For example, we can introduce a new variable like `$$cap_rootdomain`. Then combined with `$$cap_appname`, it's possible to create the subdomain:
```
$$cap_appname.$$cap_rootdomain
```

MeXaaR:
Yes exactly.

Maybe before adding this to the template system, it would be possible to add a variable to this particular app with the entire url of the app

something like this maybe
`"RUN echo 'system.url-prefix: \"$$cap_appurl\"' >> /etc/sentry/config.yml",`

githubsaturn:
If it's necessary for this app to work, I'm okay with doing it like this for now.

MeXaaR:
I'll try to make the changes on the weekend

nephthys:
I filled in some parameters for the SMTP server during the installation (the parameters are in environmental variables now) but once the application is deployed, the credentials are not memorized... How to change configuration settings after installation? 

Thanks @aldaircruz for the work anyway. ;-)

aldaircruz:
> I filled in some parameters for the SMTP server during the installation (the parameters are in environmental variables now) but once the application is deployed, the credentials are not memorized... How to change configuration settings after installation?
> 
> Thanks @aldaircruz for the work anyway. ;-)

Hi @nephthys, 

Thanks

I'm not sure why those [environment variables](https://hub.docker.com/_/sentry#environment-variable) are not working.
After the initial setup you can update those properties on the config.yml file:
https://docs.sentry.io/server/config/

githubsaturn:
FYI - `$$cap_root_domain` is now available in one-click apps (v2) as of CapRover 1.5.1

dg-otrs:
> > I filled in some parameters for the SMTP server during the installation (the parameters are in environmental variables now) but once the application is deployed, the credentials are not memorized... How to change configuration settings after installation?
> > Thanks @aldaircruz for the work anyway. ;-)
> 
> Hi @nephthys,
> 
> Thanks,
> 
> I'm not sure why those [environment variables](https://hub.docker.com/_/sentry#environment-variable) are not working.
> After the initial setup you can update those properties on the config.yml file:
> https://docs.sentry.io/server/config/

The issue seems the initial setup of the One Click App:
https://github.com/caprover/one-click-apps/blob/master/public/v2/apps/sentry.json#L75-L80

This will overwrite the ENV everytime.

After doing:
modg@captain ~ $ docker exec -it --user root 8a4c85e78b1c bash
root@sentry:/# cat /etc/sentry/config.yml
auth.allow-registration: false
beacon.anonymous: true
mail.from: ""
mail.host: ""
mail.password: ""
mail.port: 465
mail.use-tls: true
mail.username: ""
system.admin-email: ""
system.url-prefix: ""
root@sentry:/# echo '' > /etc/sentry/config.yml # make this as presistant volume, so that it wont get overwritten
root@sentry:/#

And setting up the ENV variable within the Caprover App all is good. Keep in mind to set them also for cron and worker App.

Cheers

Dennis

P.S. The system url prefix should be set correctly during the One Click App setup or not be filled empty! This occurs into the issue that the DSN is not set.

PR #64, opened 2019-05-22
Add template for Directus

incognitorobito:

I've been using looking at using the Directus CMS for a while and I thought it'd be a welcome addition to CapRover.

The Directus web app will talk to the API via a URL defined in an environment variable. Is there anyway that I can get part of this variable from CapRover? Right now as a workaround I've just defined a URL variable that the user has to fill out.

There was some talk in #60 about adding `$$cap_rootdomain` so that you can build the public facing URL `$$cap_appname.$$cap_rootdomain`. That's the URL that the Directus webapp needs.

### DIFF
--- a/public/v1/apps/directus.json
+++ b/public/v1/apps/directus.json
@@ -106,7 +106,7 @@
         "validRegex":"/^(?=.*\\d).{10,}$/"
       },
       {
-        "description":"A URL in the format of http://appname-api.rootdomain.com/_/. \n The admin web app will connect to the endpoint using this URL.",
+        "description":"A URL in the format of http://appname-api.rootdomain.com. \n The admin web app will connect to the endpoint using this URL.",
         "id":"$$cap_api_endpoint",
         "label":"Directus API Endpoint URL",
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks for the reminder about the root domain variable. I've added an issue to the main repo so we don't forget about this. https://github.com/caprover/caprover/issues/435
Since this is relatively important, I'll try to have it done for the next release.

githubsaturn:
Having said that, does the API really need to be exposed to the outer world? If not, then the app container can access it via `http://srv-captain--$$cap_appname-api`. This approach does not require the root domain. Thoughts?

incognitorobito:
The link has to be public facing so that users can use the API component of
Directus.

githubsaturn:
FYI - `$$cap_root_domain` is now available in one-click apps (v2) as of CapRover 1.5.1

incognitorobito:
I will fix the template sometime later this weekend. There's also a new update for Directus that I will roll in with the fix.

Issue #80, opened 2019-07-24
App Suggestion: Poste.io


devxpy:
Hey there!

I was trying to get [poste.io](https://poste.io/doc/) working, and saw 2 red-flags in their Web UI. 

```json
{
  "captainVersion": "1",
  "dockerCompose": {
    "version": "3.7",
    "services": {
      "$$cap_appname": {
        "image": "analogic/poste.io",
        "ports": [
          "25:25",
          "110:110",
          "143:143",
          "465:465",
          "587:587",
          "993:993",
          "995:995",
          "4190:4190"
        ],
        "volumes": [
          "/etc/localtime:/etc/localtime:ro",
          "$$cap_appname-data:/data",
          "/captain/generated/static/domains/$$cap_host/.well-known:/opt/www/.well-known."
        ],
        "restart": "always",
        "environment": {
          "HTTPS": "OFF",
          "LETSENCRYPT_EMAIL": "$$cap_email",
          "LETSENCRYPT_HOST": "$$cap_host",
          "VIRTUAL_HOST": "$$cap_host"
        }
      }
    }
  },
  "instructions": {
    "start": "SMTP + IMAP + POP3 + Antispam + Antivirus\nWeb administration + Web email\n...on your server in ~5 minutes\n\nRecommended: Use\"mail\" as the App Name.",
    "end": "Poste is deployed and available as $$cap_appname. \n\n IMPORTANT: It will take up to 2 minutes for Poste to be ready. Before that, you might see 502 error page.\n"
  },
  "variables": [
    {
      "id": "$$cap_email",
      "label": "Email Address",
      "description": "The email address used for Lets Encrpt certificates."
    },
    {
      "id": "$$cap_host",
      "label": "Hostname",
      "description": "The full hostname for this captain app. (usually mail.mydomain.com)"
    }
  ]
}
```

Could you please guide me on how I can solve these 2 issues?

![image](https://user-images.githubusercontent.com/19492893/61821282-6d051600-ae74-11e9-899e-6412f0ede961.png)

It seems like poste.io requires the `--hostname` argument to work, and caprover ignores that in docker-compose.

Not sure what the Let's encrypt warning is about.

devxpy:
Note: Mail send/receive works fine even with these warnings :)

githubsaturn:
This question would be more appropriate for poste.io.

[...]

PR #84, opened 2019-08-05
Add poste.io

devxpy:
Discussion - https://github.com/caprover/one-click-apps/issues/80


githubsaturn:
Thanks! Let's also add your findings to instructions.end, such as 
> The trick to not getting marked as spam is to properly set up the reverse DNS for your server, DKIM, and SPF records.

githubsaturn:
I accidentally merged this. But please feel free to make edits and add the docs as I suggested above.

PR #85, opened 2019-08-08
FusionAuth

r-t-k:

### DIFF
--- a/public/v1/apps/fusionauth.json
+++ b/public/v1/apps/fusionauth.json
@@ -72,7 +72,7 @@
     }
   },
   "instructions": {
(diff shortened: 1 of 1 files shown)


githubsaturn:
Hi @r-t-k 

Thanks for PR. What's the minimum spec for running this? I have a 2GB vps and this fails to start. It may be a resource issue?

r-t-k:
That’s a good question. They have this page: https://fusionauth.io/docs/v1/tech/installation-guide/system-requirements but that isn’t helpful for hardware requirements.  I’d think that 2-4gb would be enough depending on what else is eating up resources. What do your logs say?

githubsaturn:
It appears it's the elastic search that causes the problem on 2GB... It keeps restarting... Having said that, it runs fine on 4GB RAM.

Can you please add a note to `instructions.start` highlighting the minimum RAM requirement?

r-t-k:
Absolutely, thank you for helping discover this issue

PR #88, opened 2019-08-21
Fix spelling mistake

featherbear:


githubsaturn:
Thanks!

Issue #87, opened 2019-08-14
Regex on Sentry doesn't allow keys generated by Sentry


jakcharlton:
The regex is:
> ^(?> =.*\d)(?=.*[a-z])(?=.*[A-Z])[0-9a-zA-Z]{50,}$

A key generated by Sentry looks like:

docker run --rm sentry config generate-secret-key
> )%xpp06ugqbu72dt#(yjq48!&qqnptz^xjk9a9&0b^dzcu7e@+

githubsaturn:
@aldaircruz 
Would you be able to look into this please?

jakcharlton:
May also be worth checking the password regex's as both are hugely
restrictive (only letters and numbers, very long)

aldaircruz:
Hi @jakcharlton thanks for pointing this out.
@githubsaturn I will have a look into this!

githubsaturn:
Thanks to @aldaircruz this is patched. Please check again and let me know if the issue is resolved so we can close this issue.

PS: No need to update your CapRover. One-click apps dynamically get pulled from the this repo.

jakcharlton:
Much appreciated, sadly I won't have time to check in the near future, I
moved on from using sentry ... probably safe to assume it's resolved though

PR #89, opened 2019-08-23
fix(regex): update Postgres Password, Sentry Secret and Sentry Superuser Password regular expressions for validation

aldaircruz:
Hi @githubsaturn,

related with #87 

The **Sentry Secret** is using the following regex: 
https://regex101.com/r/cjfHWb/1
Should contain at least one digit, one lower case, one special character and at least 50 from the mentioned characters.

The **Postgres Password** and **Sentry Superuser Password** are using the following regex: 
https://regex101.com/r/WW64lW/1
Should contain at least one digit, one lower case, one upper case, one special character and at least 12 from the mentioned characters.

Best Regards

### DIFF
--- a/public/v2/apps/sentry.json
+++ b/public/v2/apps/sentry.json
@@ -156,14 +156,14 @@
         {
             "id": "$$cap_pg_pass",
             "label": "Postgres Password",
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #90, opened 2019-09-04
Add a Graylog to one click apps

solomon23:


uhlhosting:
@solomon23 its nice that this was added, yet is not working anymore with the latest releases, please update defaults to get the app working. @githubsaturn the app should be removed until fixed!

githubsaturn:
Thanks for the heads up. Removed.

PR #97, opened 2019-10-15
Add Countly analytics as a one-click app

vohzd:
Thank you everyone! Here is a one-click app that is confirmed to work.

This app bootstraps itself (in the same way magento and wordpress do), so there's little need to add env vars.

### DIFF
--- a/public/v2/apps/countly.json
+++ b/public/v2/apps/countly.json
@@ -1,0 +1,21 @@
+{
+    "captainVersion": "2",
+    "documentation": "Taken from https://hub.docker.com/r/countly/countly-server",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname": {
+                "image": "countly/countly-server",
+                "volumes": [
+                    "$$cap_appname-data:/var/lib/mongodb"
+                ],
+                "restart": "always",
+                "environment": {}
+            }
+        }
+    },
+    "instructions": {
+        "start": "Countly. Product Analytics for Mobile, Web, Desktop and IoT | https://count.ly",
+        "end": "Countly is deployed and available as $$cap_appname"
+    }
+}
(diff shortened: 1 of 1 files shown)


ronalddas:
@vohzd this isn't working. I think, mongodb has to be setup separately. This is the error  am getting 
```
ERROR	[db:read]	Error reading jobs {"name":"find","args":[{"status":0,"next":{"$lt":1589107956066},"name":{"$in":["api:clear","api:clearAutoTasks","api:clearTokens","api:ipcTest","api:ping","api:task","api:test","api:topEvents","api:userMerge","reports:send","push:clear","push:process","push:schedule","push:validate","server-stats:stats","assistant:generate","alerts:monitor"]}}]}
 MongoServerSelectionError: connect ECONNREFUSED 127.0.0.1:27017 {"name":"MongoServerSelectionError","reason":{"type":"Single","setName":null,"maxSetVersion":null,"maxElectionId":null,"servers":{},"stale":false,"compatible":true,"compatibilityError":null,"logicalSessionTimeoutMinutes":null,"heartbeatFrequencyMS":10000,"localThresholdMS":15,"commonWireVersion":null}}
```

githubsaturn:
If it had been tested originally and it's broken now, it's probably due to the usage of `latest`... That's why we avoid using this tag.... I'll change this to a fixed tag that was published at the time of this PR. `19.08.1`... @ronalddas  please retry with this tag `19.08.1` instead of `latest` and let me know if it works.

ronalddas:
@githubsaturn Yeah its working with `19.08.1` . Although the deployment in caprover failed with a `502` status code, but the app was created and was able to setup admin account in countly

vohzd:
Apologies! Yes when I tested with `latest` was working all OK - This is why from now on I'll be setting the explicit versions!

eborges-git:
It doesn't work recent versions. Front end runs, but server doesn't receive data. Any idea on how to install Countly recent versions (23.03.2) without having to create multiple apps?

PR #99, opened 2019-10-19
Web server One Click Apps

featherbear:
I've noticed that there aren't any one-click applications for serving websites.

I've been using basic captain-definition files, but figured that it'd be good to add them to the One-Click-Apps repository


githubsaturn:
These one click apps don't provide any useful functionality. Static apps should be deployed in standard way with the content built in.

PR #104, opened 2019-11-07
latest drone agent now requires docker socket

sneak:


sneak:
latest drone also uses a different oauth variable setup, this needs to be amended

PR #133, opened 2020-02-20
add radicale

valoricDe:

That is difficult as someone might want to add a list of users. I could build my own dockerfile but this would be unmaintained and not kept updated.
The maintainer tomsquest is only maintaining the latest tag
Hmm I dont know if I'm allowed to take the logo from the website.

### DIFF
--- a/public/v2/apps/radicale
+++ b/public/v2/apps/radicale
@@ -1,0 +1,33 @@
+{
+    "captainVersion": "2",
+    "documentation": "Taken from https://hub.docker.com/r/tomsquest/docker-radicale",
+    "displayName": "",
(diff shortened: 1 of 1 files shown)


githubsaturn:
If `latest` is the only tag that's being maintained. You'll have to fall back to sha reference of the image which is immutable. Otherwise, things can break in the future if the API changes.

Also, with regards to the icon, it's usually okay to use their icon to refer to their own product. It's similar to their name, as long as you're not stealing their name to refer to another product they should be fine with it. You can ask them on GitHub.

githubsaturn:
Closing as stale - feel free to reopen with comments addressed.

PR #150, opened 2020-04-17
Etherpad

plmorel:
Hi!

yes, though, something weird is visible through portainer I have a stopped image with img-captain-etherpad visible
- [x ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
Etherpad is rather nice regarding this, they have internal default value, so the application can be deployed with minimal choices (including no admin user)
- [x ] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
version 1.8.0 used for etherpad, 5.7 for mysql (could be 5.7.29, though 5.7 seems wiser to get security updates)
I followed the pattern used by other apps, not sure that it's enough
- [x ] Icon is added as a png file to the logos directory.
- [x ] Documentation field contains a link to a page with proper explanations on environmental variables, volumes or the docker compose file.

I choose mysql for the database side, the default dirtydb is meant for development, and there is some ongoing issue with postgres that require further setup after the default docker deployment


public/v2/apps/etherpad.json
```diff
@@ -0,0 +1,77 @@
...
+    "documentation": "https://github.com/ether/etherpad-lite/blob/develop/doc/docker.md ",
+    "displayName": "Etherpad",
+    "description": "Etherpad is a real-time collaborative editor for the web",
+    "dockerCompose": {
+        "version": "3.8",
+        "services": {
+            "$$cap_appname": {
+            "image": "etherpad/etherpad:$$cap_etherpad_version",
+            "ports": [
+                "9001:9001"
```
githubsaturn, line 12: Why is there a port mapping as well as containerHttpPort?

public/v2/apps/etherpad.json
```diff
@@ -0,0 +1,77 @@
...
+                    "MYSQL_DATABASE": "etherpad",
+                    "MYSQL_USER": "etherpad",
+                    "MYSQL_PASSWORD": "$$cap_db_pass"
+             }
+            }
+        }
+    },
+    "instructions": {
+        "start": "Etherpad is a real-time collaborative editor for the web",
+        "end": "Etherpad is a real-time collaborative editor for the web is deployed and available as srv-captain--$$cap_appname:80.  Note that the application may take some time to become available."
```
githubsaturn, line 48: `:80` is redundant.

public/v2/apps/etherpad.json
```diff
@@ -0,0 +1,77 @@
...
+            "image": "etherpad/etherpad:$$cap_etherpad_version",
+            "ports": [
+                "9001:9001"
+            ],
+            "containerHttpPort": "9001",
+            "depends_on": [
+                "$$cap_appname-db"
+            ],
+            "environment": {
+                "NODE_ENV": "$$cap_env",
```
githubsaturn, line 19: I don't think there is benefit from letting the end user set this value, we can hardcode it to `production`

public/v2/apps/etherpad.json
```diff
@@ -0,0 +1,77 @@
...
+                "DB_TYPE": "mysql",
+                "DB_HOST": "srv-captain--$$cap_appname-db",
+                "DB_PORT": "3306",
+                "DB_NAME": "etherpad",
+                "DB_USER": "etherpad",
+                "DB_PASS": "$$cap_db_pass",
+                "DB_CHARSET": "utf8mb4",
+                "TRUST_PROXY": "true"
+             }
```
githubsaturn, line 28: There is another env var in their docs, `ADMIN_PASSWORD`, do we not need it?
I am not an etherpad user, not sure if it's needed..

plmorel: If there is no admin_password set, the default is to not create an admin user.
With the choosen setup there is no authentification, so an admin user isn't extra useful
It's one less moving part

public/v2/apps/etherpad.json
plmorel: a brain fart, sorry

public/v2/apps/etherpad.json
plmorel: done

public/v2/apps/etherpad.json
plmorel: removed

githubsaturn:
Great, thanks! You can now test in production.

PR #156, opened 2020-04-22
Jitsi

chrisft25:

- [☑️ ] I have tested the template using the method described in README.md thoroughly
- [☑️ ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [☑️ ] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [☑️ ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [☑️ ] Icon is added as a png file to the logos directory.
- [☑️ ] Documentation field contains a link to a page with proper explanations on environmental variables, volumes or the docker compose file.


public/v2/apps/jitsi.json
```diff
@@ -0,0 +1,255 @@
...
+    "documentation": "https://github.com/jitsi/docker-jitsi-meet",
+    "displayName": "Jitsi",
+    "description": "Free open-source video conferencing for web & mobile",
+    "dockerCompose": {
+      "version": "3",
+      "services": {
+        "$$cap_appname-web": {
+          "image": "jitsi/web:$$cap_tag_version",
+          "restart": "unless-stopped",
+          "ports": ["$$cap_https_port:443", "$$cap_http_port:80"],
```
githubsaturn, line 12: we shouldn't map the http(s) ports. Why don't we use the `containerHttpPort` ?

public/v2/apps/jitsi.json
```diff
@@ -0,0 +1,255 @@
...
+    "instructions": {
+      "end": "Containers were created successfully. You should wait for a few minutes before enter to Jitsi Web.",
+      "start": "This one click app is based on Jitsi docker image."
+    },
+    "variables": [
+        {
+            "label": "Docker tag version",
+            "description": "You can check the version in Docker Hub: https://hub.docker.com/r/jitsi/web/tags",
+            "defaultValue": "4416",
+            "id": "$$cap_tag_version"
```
githubsaturn, line 215: Are all services using the same docker image tag?

chrisft25:
Hi, yes, you can try it on my server: https://jitsi-web.server.chrisfuentes.me:8443/

I have been looking for the official minimum requirements and I didn't find it. I found this tutorial who says the following:

2GHz CPU

1 GB Memory

25 GB Disk

10 GbE Network




Source: https://lambda-it.ch/blog/post/setup-jitsi-meeting-server

public/v2/apps/jitsi.json
chrisft25: Yes, you can check it all here:

https://hub.docker.com/r/jitsi/web/tags
https://hub.docker.com/r/jitsi/jvb/tags
https://hub.docker.com/r/jitsi/web/tags

public/v2/apps/jitsi.json
chrisft25: How does it work? I have been trying and I couldn't make it work with that property

githubsaturn: Port 443 without certs doesn't provide a valid HTTPS connection. You probably have to use port `80`. So simply remove this `ports` section. You don't even need to specify `containerHttpPort` as the default value is already 80. So the network traffic will be
```
User === HTTPS ===>> nginx (CapRover) : port 433 === http ===>> jitsi : port 80
```

githubsaturn:
I still don't understand why port mappings are required. But regardless, if this is tested and working, we can merge it to have a starting point, we can improve later. Thanks :+1:

monkeydust:
Hi, firstly thanks for contribution. I wanted to know how one can secure the jitsi instance deployed through caprover. Right now anyone with URL can start video calls. The concept exists here - https://jitsi.github.io/handbook/docs/devops-guide/secure-domain. 

Any way / guidance on how to deploy this?

Thanks

PR #157, opened 2020-04-22
Add Bitwarden_rs

vohzd:
Hello, added Bitwarden_rs & tested.


public/v2/apps/bitwardenrs.json
```diff
@@ -0,0 +1,33 @@
...
+    "captainVersion": "2",
+    "documentation": "Taken from https://hub.docker.com/r/bitwardenrs/server",
+    "displayName": "Bitwarden_rs",
+    "description": "Lightweight fully featured Rust implementation of Bitwarden.",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname": {
+                "dockerfileLines": [
+                    "FROM bitwardenrs/server:$$cap_bitwardenrs_version"
```
githubsaturn, line 11: If this is the only line here, we should simply use `imageName` property.

vohzd: Sure, added here;
https://github.com/caprover/one-click-apps/pull/157/commits/cf8029ad4380eb2da16fa860eb5fc73c2d98c219

githubsaturn:
Thank you

PR #165, opened 2020-05-02
passbolt

belcboo:

This app uses the official image and documentation from Passbolt. Is another alternative to password manager for users and teams.


githubsaturn:
Have you fully tested this? Some of the keys set here are ignored by CapRover. If this is tested and working, it means that those keys are not important at all. For example `command` is not used by CapRover.

belcboo:
> Have you fully tested this? Some of the keys set here are ignored by CapRover. If this is tested and working, it means that those keys are not important at all. For example `command` is not used by CapRover.

Yes, I’m currently using it in a project with some friends and we haven’t had any issue. If you want me to remove those keys I can, just let me know what other key is not used by Caprover. 

Thanks.

githubsaturn:
The info is in README:
image

In your case,  you can remove `tty` `tmpfs` and `command`

githubsaturn:
Great, thanks! Once it's available on the official repo, I suggest test it again to make sure that it works as expected!

PR #173, opened 2020-05-22
Feature add mautic one app deployment

LiquidITGuy:
few comments about this MR : 
- tested and deployed in few second on my server so I dont wrote about waiting time
- updated version of mautic on docker are only not fixed (latest, fpm, apache) and the fixed beta. so to prevent build failed later I chose the beta one.


LiquidITGuy:
Capture d’écran 2020-05-22 à 22 01 44
Capture d’écran 2020-05-22 à 22 01 57
Capture d’écran 2020-05-22 à 22 02 04

TechGuyTN:
Thank you @LiquidITGuy :D :D :D

LiquidITGuy:
with pleasure @TechGuyTN ;)

public/v2/apps/mautic.json
```diff
@@ -0,0 +1,77 @@
...
+    "services": {
+      "$$cap_appname-db": {
+        "documentation": "Taken from https://hub.docker.com/_/percona",
+        "image": "percona/percona-server:$$cap_percona_version",
+        "notExposeAsWebApp": "true",
+        "volumes": [
+          "$$cap_appname-db-data:/var/lib/mysql"
+        ],
+        "ports": [
+          "3306:3306"
```
githubsaturn, line 16: You don't need to expose the port to the outside world. You can remove this.

public/v2/apps/mautic.json
```diff
@@ -0,0 +1,77 @@
...
+    {
+      "id": "$$cap_db_pass",
+      "label": "Database password",
+      "description": "",
+      "validRegex": "/.{1,}/"
+    },
+    {
+      "id": "$$cap_mautic_version",
+      "label": "Mautic Version",
+      "defaultValue": "v3.0.0-beta",
```
githubsaturn, line 65: It appears that v3 is not a stable version yet, we should use v2.x, perhaps `2.13-apache`

LiquidITGuy: I think the 2.13 apache is even more buggy (last real production version [apache] has been updated 2 month ago, but last fixed version (excluding beta one) was not updated since 2 years)
I added beta into the name to warn about the beta version and do not want to use a not fixed version.

public/v2/apps/mautic.json
LiquidITGuy: agreed, solved

public/v2/apps/mautic.json
LiquidITGuy: information message is added on the description

LiquidITGuy:
tried after reviews changes and it still work. :)

PR #174, opened 2020-05-22
Update nextcloud.json to upgrade nextcloud

LiquidITGuy:
upgrade nextcloud from 15 to 18.0.4
from no config to fully configure to start

### DIFF
--- a/public/v2/apps/nextcloud.json
+++ b/public/v2/apps/nextcloud.json
@@ -1,31 +1,83 @@
 {
     "captainVersion": "2",
     "documentation": "Taken from https://hub.docker.com/_/nextcloud",
-    "displayName": "",
+    "displayName": "nextcloud",
     "description": "Nextcloud is a suite of client-server software for creating and using file hosting services",
     "dockerCompose": {
         "version": "3.3",
         "services": {
-            "$$cap_appname": {
-                "image": "nextcloud:$$cap_nextcloud_version",
+            "$$cap_appname-db": {
+                "documentation": "Taken from https://hub.docker.com/_/mariadb",
+                "image": "mariadb:$$cap_mariadb_version",
+                "notExposeAsWebApp": "true",
                 "volumes": [
-                    "$$cap_appname-data:/var/www/html"
+                    "$$cap_appname-db-data:/var/lib/mysql"
                 ],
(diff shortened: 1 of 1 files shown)


LiquidITGuy:
Capture d’écran 2020-05-23 à 00 33 54
Capture d’écran 2020-05-23 à 00 34 21

public/v2/apps/nextcloud.json
```diff
@@ -1,31 +1,83 @@
...
+                    "$$cap_appname-db"
+                ],
+                "documentation": "Taken from https://hub.docker.com/_/nextcloud",
                 "image": "nextcloud:$$cap_nextcloud_version",
                 "volumes": [
-                    "$$cap_appname-data:/var/www/html"
+                    "$$cap_appname-website:/var/www/html",
+                    "$$cap_appname-apps:/var/www/html/custom_apps",
+                    "$$cap_appname-config:/var/www/html/config",
+                    "$$cap_appname-data:/var/www/html/data"
```
githubsaturn, line 34: You don't need to define the last 3 volumes. Only the top one would be enough.

public/v2/apps/nextcloud.json
```diff
@@ -1,31 +1,83 @@
...
+        "description": "Username for the database using mysql."
+    },{
+        "id": "$$cap_admin_user",
+        "label": "admin name",
+        "defaultValue": "admin",
+        "description": "Name of the Nextcloud admin user."
+    },{
+        "id": "$$cap_admin_pass",
+        "label": "admin password",
+        "description": "Password for the Nextcloud admin user."
```
githubsaturn, line 81: Add valid regex for all params here, otherwise user can skip them and the installation would fail.

public/v2/apps/nextcloud.json
LiquidITGuy: agreed, fixed

public/v2/apps/nextcloud.json
LiquidITGuy: I was thinking all params was required, thanks for the comment, I will add it on all my pr :)
fixed

LiquidITGuy:
fixed and test with new version

PR #175, opened 2020-05-23
Feature add limesurvey

LiquidITGuy:
few words about this MR : 
- there is no official docker version of Limesurvey but I take the most popular (more than 1M download)
- this docker version has a mysql database inside but is not settable by env var and retrieve informations is hard for the user (it need to retrieve log). As recommended on the documentation I made a second database to connect Limesurvey


LiquidITGuy:
Capture d’écran 2020-05-23 à 22 35 29
Capture d’écran 2020-05-23 à 22 36 26
Capture d’écran 2020-05-23 à 22 36 30

public/v2/apps/limesurvey.json
```diff
@@ -0,0 +1,69 @@
...
+    },{
+      "id": "$$cap_db_name",
+      "label": "database name",
+      "defaultValue": "limesurvey",
+      "description": "name for the database."
+    },{
+      "id": "$$cap_db_user",
+      "label": "database user",
+      "defaultValue": "limesurvey",
+      "description": "Username for the database using mysql."
```
githubsaturn, line 66: valid regex is missing from multiple params. Please add them.

LiquidITGuy: done

LiquidITGuy:
not tried again after dev fixes (only regex changes)

PR #176, opened 2020-05-23
Feature add n8n.io

LiquidITGuy:
few words about this MR : 
- https does not work (even with properly configured docker image (the clients works, but the api seems unreachable)
- the service uses basic web auth to autenticate user and it is removable. I tried to not use the auth from n8n but the one from caprover.
n8n service provides api creation for webhook that are secured from other way, activate caprover securisation will block the webhook function.
Second reason : if we activate the classical authent from n8n, with 2 web auth, we got some problems.

I would love to add notes to explain it more but there is no place so I put it all into the end part.
ps : to try trigger, dont forget to start the trigger listening on test or activate in production


LiquidITGuy:
2nd précision : it will work in https but with an error (connexion lost) but it seems to work very well (issue should be created on n8n side)

LiquidITGuy:
Capture d’écran 2020-05-24 à 01 09 07
Capture d’écran 2020-05-24 à 01 09 38
on http:
Capture d’écran 2020-05-24 à 01 09 51
on https : 
Capture d’écran 2020-05-24 à 01 10 06

public/v2/apps/n8n-io.json
```diff
@@ -0,0 +1,94 @@
...
+      "description": "Timezone used by the server. Please check out this page for valid time"
+    },{
+      "id": "$$cap_user",
+      "label": "user",
+      "defaultValue": "root",
+      "description": "user you need to use to login into the web interface"
+    },{
+      "id": "$$cap_pass",
+      "label": "password",
+      "description": "password you need to use to login into the web interface"
```
githubsaturn, line 91: valid regex needs to be added to all params here.

LiquidITGuy: done

PR #187, opened 2020-06-14
Add Shopware

shyim:
This adds Shopware 6 as one click app :). The image itself is also maintained from me. I am an employee from Shopware :) 


public/v2/apps/shopware.json
```diff
@@ -0,0 +1,117 @@
+{
+    "captainVersion": "2",
+    "documentation": "See https://github.com/shyim/shopware-image for tags and other variables",
+    "displayName": "Shopware",
+    "description": "Shopware is a trendsetting ecommerce platform to power your online business. Our ecommerce solution offers the perfect combination of beauty & brains you need to build and customize a fully responsive online store.",
```
githubsaturn, line 5: I think the description is a bit too long. It'll mess up with the UI. I suggest removing `Our ecommerce solution...` and add it to `instructions.start`.

public/v2/apps/shopware.json
```diff
@@ -0,0 +1,117 @@
...
+            "id": "$$cap_db_user",
+            "label": "Database user",
+            "defaultValue": "shopwareuser",
+            "validRegex": "/^([a-zA-Z0-9])+$/"
+        },
+        {
+            "id": "$$cap_db_pass",
+            "label": "Database password",
+            "description": "",
+            "validRegex": "/.{1,}/"
```
githubsaturn, line 83: You probably want to limit this to alpha-numeric values - because otherwise your password in your database URL will not work properly.

public/v2/apps/shopware.json
```diff
@@ -0,0 +1,117 @@
...
+                    "$$cap_appname-sitemap:/var/www/html/public/sitemap",
+                    "$$cap_appname-thumbnail:/var/www/html/public/thumbnail",
+                    "$$cap_appname-jwt:/var/www/html/config/jwt"
+                ]
+            }
+        }
+    },
+    "instructions": {
+        "start": "Shopware is a trendsetting ecommerce platform to power your online business.",
+        "end": "Shopware is successfully deployed. Please wait a minute before accessing the shop."
```
githubsaturn, line 60: If you are an employee of Shopware and would like to maintain this app, feel free to add your github URl to the instructions.end so that people know how to ask for support.

shyim: Done :)

public/v2/apps/shopware.json
shyim: I did copycat that from wordpress (https://github.com/caprover/one-click-apps/blob/master/public/v2/apps/wordpress.json#L57)
Should I create another PR? 🙈

PR #195, opened 2020-06-23
Validate json split

LiquidITGuy:

- [v] I am not adding a new template

### DIFF
--- a/validate_json.js
+++ b/validate_json.js
@@ -1,4 +1,5 @@
 /*jshint esversion: 6 */
+// Requirements
 const path = require('path')
 const fs = require('fs-extra')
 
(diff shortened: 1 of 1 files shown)


validate_json.js
```diff
@@ -1,71 +1,73 @@
...
- Promise.resolve()
-     .then(function () {
-         return validate()
-     })
-     .catch(function (err) {
-         console.error(err)
-         process.exit(127)
-     })
\ No newline at end of file
+ async function validate() {
```
githubsaturn, line 38: FTR - we don't use async/await in our front-end and backend code. But this is just a build code, so it's fine.

validate_json.js
```diff
@@ -1,71 +1,73 @@
...
+const isLogoValid = ({ existsSync, statSync }, logoFullPath) => existsSync(logoFullPath) && statSync(logoFullPath).isFile()
+const isVersionValid = (applicationVersion, capVersion) => applicationVersion === capVersion
+
+// Displayed Messages
+const printablePath = fullPath => fullPath.substr(fullPath.indexOf(`/${PUBLIC}`))
+const versionErrorMessage = (appVersion, capVersion, filename) => `unmatched versions ${appVersion} ${capVersion} for ${filename}`
+const logoErrorMessage = (filename, logoFullPath) =>`Cannot find logo for ${filename} ${printablePath(logoFullPath)}`
+const defaultDisplayName = appName => appName.charAt(0).toUpperCase() + appName.slice(1)
+const endMessage = appNumber => `${appNumber} applications validated.`
+const validatedMessage = filename => `Validated ${filename}`
```
githubsaturn, line 24: Two notes here:
- Some of these method names are variable names, method names usually have a verb. 
- Extraction to a method is good if it's repeating itself. Most of these methods (if not all of them) are only used once. Inlining them will make the code much easier to read.

validate_json.js
```diff
@@ -1,71 +1,73 @@
...
+
+const DISPLAY = {
```
githubsaturn, line 26: Based on my comment above, this is not needed. But if it was, we should have gone with better naming:
```
DISPLAY -> Logger

ERROR  -> logError
...
```

validate_json.js
```diff
@@ -1,71 +1,73 @@
...
+const PUBLIC = 'public'
+const pathOfPublic = path.join(__dirname, PUBLIC)
+const VERSION = 2
+const LOGO_EXTENSION = '.png'
+const ONE_APP_EXTENSION = '.json'
+const APP_FOLDER = 'apps'
+const LOGO_FOLDER = 'logos'
+
+// VALIDATION functions
+const isLogoValid = ({ existsSync, statSync }, logoFullPath) => existsSync(logoFullPath) && statSync(logoFullPath).isFile()
```
githubsaturn, line 15: Don't pass `fs` as a parameter to a method.

LiquidITGuy:
Json are not used anymore

PR #205, opened 2020-07-01
Added a one-click Eclipse Mosquitto MQTT Broker App

Nauman3S:


Nauman3S:
Eclipse MQTT Broker

public/v2/apps/eclipse-mosquitto.json
```diff
@@ -0,0 +1,60 @@
...
+            "image": "eclipse-mosquitto:$$cap_eclipse_version",
+            "volumes": [
+                "$$cap_appname-eclipse-config:/mosquitto/config/",
+                "$$cap_appname-eclipse-data:/mosquitto/data/",
+                "$$cap_appname-eclipse-log:/mosquitto/log/"
+            ],
+            "notExposeAsWebApp": "true",
+            "ports": [
+                    "$$cap_mqtt_port:1883",
+                    "$$cap_ws_port:9001"],
```
githubsaturn: These port don't have to be mapped to host ports necessarily if the app is only being used insternally via other apps hosted on the same CapRover instance. 

Let's remove them.

public/v2/apps/eclipse-mosquitto.json
```diff
@@ -0,0 +1,60 @@
...
+        "restart": "always"
+  }
+}
+},
+
+"instructions": {
+    "start": "Eclipse Mosquitto is an open source message broker which implements MQTT version 5, 3.1.1 and 3.1. Enter Parameters below",
+    "end": "Eclipes-Mosquitto is deployed and available as srv-captain--$$cap_appname:$$cap_mqtt_port to other apps and srv-captain--$$cap_appname:$$cap_ws_port as a websocket."
```
githubsaturn: Are you sure you tested these and it worked? `srv-captain--$$cap_appname:$$cap_mqtt_port`  and `srv-captain--$$cap_appname:$$cap_ws_port` are incorrect.
As mentioned in the other comments these are the host ports. When you're using `srv-captain--*****` you are binding to the container ports. There for these two need to be changed to
```
srv-captain--$$cap_appname:1883 [mqtt port]
srv-captain--$$cap_appname:9001 [websocket]
```

i.e. when you are binding locally to these services, the ports are always 1883 and 9001. See other apps, for example [MySQL](https://github.com/caprover/one-click-apps/blob/master/public/v2/apps/mysql.json#L22) for more details.

PR #206, opened 2020-07-02
Update prisma.json to Prisma 1

Jolg42:
Because the new default is Prisma 2 see https://github.com/prisma/prisma

Minimal change :)


githubsaturn:
Thanks!

Issue #72, opened 2019-06-16
App Suggestion: [dolibarr]


oda-alexandre:
**dolibarr**
- site web : https://www.dolibarr.org/
- docker exemple : https://hub.docker.com/r/alexandreoda/dolibarr

thank you for your work

jacquant:
Hello,
```yaml
{
    "captainVersion": "2",
    "documentation": "https://github.com/caprover/one-click-apps",
    "dockerCompose": {
        "version": "3.8",
        "services": {
            "$$cap_appname-db": {
                "image": "postgres:$$cap_postgres-version",
                "notExposeAsWebApp": "true",
                "volumes": [
                    "$$cap_appname-db:/var/lib/postgresql/data"
                ],
                "restart": "always",
                "environment": {
                    "POSTGRES_USER": "dolibarr",
                    "POSTGRES_PASSWORD": "$$cap_pg-pass",
                    "POSTGRES_DB": "dolibarr"
                }
            },
            "$$cap_appname": {
                "image": "upshift/dolibarr:$$cap_dolibarr-version",
                "containerHttpPort": "80",
                "depends_on": [
                    "$$cap_appname-db"
                ],
                "volumes": [
        			"$$cap_appname-html:/var/www/html",
        			"$$cap_appname-docs:/var/www/documents"
                ],
                "restart": "always",
                "environment": {
                    "DOLI_DB_HOST": "srv-captain--$$cap_appname-db",
                    "DOLI_DB_NAME": "dolibarr",
                    "DOLI_DB_ROOT_LOGIN": "dolibarr",
                    "DOLI_DB_ROOT_PASSWOORD": "$$cap_pg-pass",
                    "DOLI_DB_USER": "dolibarr",
                    "DOLI_DB_PORT": 5432,
                    "DOLI_DB_TYPE": "pgsql",
                    "DOLI_DB_PASSWORD": "$$cap_pg-pass",
                    "DOLI_ADMIN_LOGIN": "$$cap_admin-login",
                    "DOLI_ADMIN_PASSWORD": "$$cap_admin-pass",
                    "DOLI_URL_ROOT": "http://$$cap_appname.$$cap_root_domain"
                }
            }
        },
        "volumes":{
		"$$cap_appname-html":{},
		"$$cap_appname-docs":{},
 		"$$cap_appname-db": {}
        }
    },
    "instructions": {
        "start": "Dolibarr is an Open Source ERP & CRM for Business",
[...]

PR #208, opened 2020-07-03
Dolibarr One-Click-Apps

jacquant:

Closes https://github.com/caprover/one-click-apps/issues/72


public/v2/apps/dolibarr.json
```diff
@@ -0,0 +1,89 @@
+{
+    "captainVersion": "2",
+    "documentation": "https://github.com/caprover/one-click-apps",
```
githubsaturn, line 3: Please link the documentation where the docker compose was taken from so that in the future if we need to change/add any more env variable or volumes, we know where to look.

githubsaturn:
Thank you @jacquant !

PR #212, opened 2020-07-05
Odoo One Click App

jacquant:


public/v2/apps/odoo.json
```diff
@@ -0,0 +1,70 @@
+{
+    "captainVersion": "2",
+    "documentation": "https://hub.docker.com/_/odoo?tab=description",
+    "description": "Odoo, formerly known as OpenERP, is a suite of open-source business apps written in Python and released under the AGPL license. This suite of applications covers all business needs, from Website/Ecommerce down to manufacturing, inventory and accounting, all seamlessly integrated. It is the first time ever a software editor managed to reach such a functional coverage. Odoo is the most installed business software in the world. Odoo is used by 2.000.000 users worldwide ranging from very small companies (1 user) to very large ones (300 000 users)",
```
githubsaturn, line 4: Description is too long and messes up with the UI - let's reduce it to 150 char or less. Move the extra parts to instructions.start.

public/v2/apps/odoo.json
```diff
@@ -0,0 +1,70 @@
...
+        "db": {
+            "driver": "bridge"
+        },
+        "search": {
+            "driver": "bridge"
+        }
+    },
+    "instructions": {
+        "start": "Odoo is an open source product CRM",
+        "end": "Odoo is deployed and available as srv-captain--$$cap_appname:8069 to other apps."
```
githubsaturn, line 46: Is Odoo an application that will be used by other applications on CapRover or it has a UI that user will use directly?

If it's got a UI of its own, there is no need in using `srv-captain--$$cap_appname:8069`, instead just just 
```
You can access the dashboard from http://$$cap_appname.$$cap_root_domain
```

public/v2/apps/odoo.json
```diff
@@ -0,0 +1,70 @@
...
+                }
+            },
+            "$$cap_appname": {
+                "image": "odoo:$$cap_odoo_version",
+                "containerHttpPort": "8069",
+                "depends_on": ["$$cap_appname-db"],
+                "volumes": [
+                    "odoo-web-data:/var/lib/odoo",
+                    "config:/etc/odoo",
+                    "addons:/mnt/extra-addons"
```
githubsaturn, line 26: append `$$cap_appname` to these volume names, otherwise, the second instance will use the same volumes.

PR #214, opened 2020-07-05
Openldap One Click App

jacquant:


public/v2/apps/phpldapadmin.json
```diff
@@ -0,0 +1,77 @@
...
+                }
+            },
+            "php$$cap_appnameadmin": {
+                "image": "osixia/phpldapadmin:$$cap_phpldapadmin_version",
+                "containerHttpPort": "443",
+                "restart": "always",
+                "depends_on": ["$$cap_appname"],
+                "environment": {
+                    "PHPLDAPADMIN_TRUST_PROXY_SSL": true,
+                    "PHPLDAPADMIN_LDAP_HOSTS": "$$cap_appname.$$cap_root_domain",
```
githubsaturn, line 26: Wouldn't `srv-captain--$$cap_appname` work too? If it works it'll be a better alternative as it uses an internal connection.

public/v2/apps/phpldapadmin.json
```diff
@@ -0,0 +1,77 @@
...
+    "documentation": "https://github.com/osixia/docker-openldap",
+    "description": "OpenLDAP with administration interface",
+    "displayName": "OpenLDAP + phpLDAPadmin",
+    "dockerCompose": {
+        "version": "3.8",
+        "services": {
+            "$$cap_appname": {
+                "image": "osixia/openldap:$$cap_openldap_version",
+                "restart": "always",
+                "ports": ["389:389", "636:636"],
```
githubsaturn, line 12: Do ports really need to be exposed? If it's for internal use, you can access this by `srv-captain--$$cap_appname:389` without port mapping.

public/v2/apps/phpldapadmin.json
```diff
@@ -0,0 +1,77 @@
...
+        {
+            "id": "$$cap_openldap_ldap-organisation",
+            "label": "Ldap Organisation",
+            "description": "The Organization name",
+            "validRegex": "/.{1,}/"
+        },
+        {
+            "id": "$$cap_openldap_ldap-domain",
+            "label": "Ldap Host",
+            "description": "The host",
```
githubsaturn, line 54: Maybe provide an example in the description as the regex rule is quite complex to understand. Something like `myhost.domain.something` would work, right?

jacquant: Ok I'll do that! I take the advice into accounts for the next PRs :smile:

public/v2/apps/phpldapadmin.json
jacquant: In my case, I needed it to be accessible from other machines. I can remove it and explain how to activate it in the end instructions.

githubsaturn: That'd be perfect! Thanks!

public/v2/apps/phpldapadmin.json
jacquant: It works so I change :)

public/v2/apps/phpldapadmin.json
```diff
@@ -0,0 +1,72 @@
...
+            "$$cap_appname": {
+                "image": "osixia/openldap:$$cap_openldap_version",
+                "restart": "always",
+                "notExposeAsWebApp": "true",
+                "environment": {
+                    "LDAP_ORGANISATION": "$$cap_openldap_ldap-organisation",
+                    "LDAP_DOMAIN": "$$cap_root_domain",
+                    "LDAP_ADMIN_PASSWORD": "$$cap_openldap_ldap-admin-password"
+                }
+            },
```
githubsaturn, line 18: One last question - doesn't openldap require any persistent volume? With current implementation it'll lose all its data upon restart.

jacquant: Done :smile:

githubsaturn:
Superb! Thanks!

Issue #219, opened 2020-07-07
App Suggestion: OpenVPN-AS


olla-dev:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
Have any open source vpn server like openvpn-as

**Do you know if there is any official Docker image for the app?**
there is a docker image from linuxserver (linuxserver/openvpn-as/)

tomholford:
@m4200 FYI I have started a PR for this template, since I could also use a VPN :) Will update here when it's ready for review.

tomholford:
update: per this comment from @githubsaturn, this app is not currently compatible with caprover

https://github.com/caprover/one-click-apps/pull/222#issuecomment-656083740

dramikei:
Hi, sorry to bump an old issue. But it turns out this issue couldn't be resolved due to docker swarm not supporting capabilities at the time ( #222 )

Capabilities support was added for swarm in [Docker 20.10.0](https://github.com/docker/cli/pull/2687) which should allow us to port openvpn-as or even wireguard to Caprover.

But before that we will need to add support for caprover to parse `cap_add` parameter of docker-compose.yml, which probably is ignored currently as stated in `README.md`:

```
## Services

- Other than image, environment, ports, volumes, depends_on, and hostname, other parameters are currently being ignored by CapRover. If you need a particular parameter, please file an issue, and we'll add it to the respected list.
```

dramikei:
Opened an issue in caprover repo regarding this: caprover/caprover#1100

prakasharyaman:
Any updates on openVpn on Caprover ?

PR #222, opened 2020-07-09
[WIP] OpenVPN-as

tomholford:
This adds a one-click template for [OpenVPN Access Server](https://openvpn.net/index.php/access-server/overview.html), using [this Docker image](https://hub.docker.com/r/linuxserver/openvpn-as/).

This addresses the request in #219.


githubsaturn:
I doubt that this will work. It requires `--cap-add=NET_ADMIN` which is unavailable in Docker swarm mode: https://github.com/moby/moby/issues/25885

tomholford:
@githubsaturn Ah, thanks for the heads up. Glad I didn't spend any more time on this!

tomholford:
@githubsaturn Do you think something like this would work with CapRover? Seems like there is a workaround to launch a `--privileged` container within a swarm

https://alex.thom.ae/2020/02/22/openvpn-server-docker-swarm/

githubsaturn:
Yea this is a pretty clever idea!

It's quite hacky, so it requires some time just experimenting and tweaking things to make this work properly. But other than that, the approach looks good!

tomholford:
@githubsaturn Thanks for following up. Admittedly, this kind of thing is out of my area of expertise. But maybe if I have time in the next few weeks, I will give it a shot

PR #223, opened 2020-07-09
Creating Joomla template

chrisft25:


public/v2/apps/joomla.json
```diff
@@ -0,0 +1,58 @@
+{
+    "captainVersion": "2",
+    "documentation": "Taken from https://hub.docker.com/_/joomla?tab=description. Based on Wordpress - No Database template.",
+    "displayName": "Joomla - No Database",
+    "description": "This will create a Joomla only. You will need to create and configure the database information manually. Intended for advanced users.",
+    "dockerCompose": {
+        "version": "3.3",
+        "services": {
+            "$$cap_appname-joomla": {
+                "image": "joomla:latest",
```
githubsaturn, line 10: image

Please avoid using latest tag, and use `$$cap_joomla_version` that you created below. This is part of the checklist that you checked.

chrisft25: You're right! I forgot to add the variable to the image. I have done it already!

githubsaturn: @chrisft25 - thanks! But the variable's default value is still set to `latest` - we have to change this to a fixed value.

chrisft25: It's done!

githubsaturn:
Great, thanks!

PR #233, opened 2020-07-22
added option to select database, mariadb or mysql

ethancedwards8:
Adding option to use either mariadb or mysql database. Doesn't change defaults.

### DIFF
--- a/public/v2/apps/wordpress.json
+++ b/public/v2/apps/wordpress.json
@@ -7,7 +7,7 @@
         "version": "3.3",
         "services": {
             "$$cap_appname-db": {
-                "image": "mysql:$$cap_mysql_version",
+                "image": "$$cap_db_type:$$cap_mysql_version",
                 "notExposeAsWebApp": "true",
                 "volumes": ["$$cap_appname-db-data:/var/lib/mysql"],
                 "restart": "always",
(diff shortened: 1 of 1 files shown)


public/v2/apps/wordpress.json
```diff
@@ -59,10 +59,17 @@
             "validRegex": "/^([^\\s^\\/])+$/"
         },
         {
-            "id": "$$cap_mysql_version",
-            "label": "MySQL Version",
+            "id": "$$cap_db_type",
+            "label": "Database Type",
+            "defaultValue": "mysql",
+            "description": "You can either choose mariadb or mysql, you need to change the version according to which DB is selected. It is case sensitive.",
+            "validRegex": "/.{1,}/"
```
githubsaturn, line 66: Let's change the regex to only allow mysql and mariadb

ethancedwards8: I'm not very in regex, any clues on how to do that?

githubsaturn: Maybe something like this?
```
"/^(mysql|mariadb)$/"
```

ethancedwards8: I'll try it out.

ethancedwards8:
That seems to have worked

public/v2/apps/wordpress.json
ethancedwards8: That worked

PR #234, opened 2020-07-22
added cron container, many apps like the news app will not function without a cron container

ethancedwards8:
Title


githubsaturn:
This doesn't break the normal functionality, right? NextCloud is a widely used one-click app, just wanted to be sure.

ethancedwards8:
Nope.

githubsaturn:
Thanks!

PR #236, opened 2020-07-26
MVP: Elastic stack security improvements

sergioisidoro:
So... in light of the recent meow attacks - https://arstechnica.com/information-technology/2020/07/more-than-1000-databases-have-been-nuked-by-mystery-meow-attack/

> At the time this post went live, the Shodan computer search site showed that 987 ElasticSearch .. had been nuked by Meow

I was wondering if having the passwords as env variables is worse than having unsecured elastic instances .

**This is a suggestion**. I tried using caprover's http basic auth on elastic, but it doesn't play very well with the rest of the elasticsearch stack (eg. I cannot connect to Kibana, even with `https://user:pass@elasticsearch` as the instance url)

If you feel like this is an edge case, feel free to close this PR

### DIFF
--- a/public/v2/apps/elasticsearch.json
+++ b/public/v2/apps/elasticsearch.json
@@ -20,7 +20,10 @@
                     "node.name": "$$cap_appname",
                     "node.master": "$$cap_elasticsearch_node_master",
                     "node.data": "$$cap_elasticsearch_node_data",
-                    "discovery.seed_hosts": "$$cap_elasticsearch_discovery_seed_hosts"
+                    "discovery.seed_hosts": "$$cap_elasticsearch_discovery_seed_hosts",
+                    "xpack.security.enabled":  "$$cap_elasticsearch_enable_security",
+                    "xpack.security.transport.ssl.enabled": "$$cap_elasticsearch_enable_security",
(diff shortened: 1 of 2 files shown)


sergioisidoro:
Apparently this works if we setup the nginx basic auth on the elastic search instance, and use the `ELASTICSEARCH_USERNAME` and `ELASTICSEARCH_PASSWORD` in the kibana side. 

So these changes to the elasticsearch service are not entirely needed.

githubsaturn:
Not sure if this answers your question, but here it is:
You probably don't even need to expose your elasticsearch to the public (behind nginx). In that case, you can check `Do not expose as webapp` then you wouldn't even need to password protect this. Instead, your other caprover apps (like Kibana) can access this app via `http://srv-captain--yourelasticsearchname`

sergioisidoro:
In my case I do need to expose it since some of my elastic clients are outside the docker network.

If you think the most common case is to have elastic not publicly available then we can close this :)

githubsaturn:
Yea - let's close this PR. Feel free to open a PR addressing these points in instructions.end text. Users can then choose to apply password if they need to.

PR #240, opened 2020-08-09
Fix broken deployments for Chatwoot

pranavrajs:
Use v1.6.3 as the default image.

### Known issues for 1.6.3

- Seed data won't be populated in the DB, they would need to create a new account instead of default account. This would be fixed in v1.7.0. For now, I've updated the instructions.


githubsaturn:
Thanks! Just performed a test and it seems to be working.

PR #241, opened 2020-08-09
MongoDB backup Amazon S3

felipe-augusto:


felipe-augusto:
the docker image only has the [`latest` tag](https://hub.docker.com/r/halvves/mongodb-backup-s3/tags), maybe should try using the digest?

githubsaturn:
Yes, please use a sha tag.

public/v2/apps/mongodb-backup-s3.json
```diff
@@ -0,0 +1,134 @@
...
+            "id": "$$cap_backup_folder",
+            "defaultValue": "mongo_dumps/",
+            "label": "BACKUP_FOLDER"
+        },
+        {
+            "description": "The host/ip of your mongodb database",
+            "id": "$$cap_mongodb_host",
+            "label": "MONGODB HOST",
+            "validRegex": "/^.+$/",
+            "defaultValue": "localhost"
```
githubsaturn, line 67: This default value makes is easy for the user to just click on deploy and miss the fact that this is not right. Maybe remove the default value here, and just add some bit to the description. Like:
```
            "description": "The host/ip of your mongodb database. For CapRover hosted database, use srv-captain--your-mongodb-name",
```

public/v2/apps/mongodb-backup-s3.json
```diff
@@ -0,0 +1,134 @@
...
+            "label": "EXTRA_OPTS"
+        },
+        {
+            "description": "The interval of cron job to run mongodump. 0 3 * * * by default, which is every day at 03:00hrs.",
+            "id": "$$cap_cron_time",
+            "label": "CRON_TIME",
+            "defaultValue": "0 3 * * *"
+        },
+        {
+            "description": "timezone",
```
githubsaturn, line 104: What's the difference between timezone and cron timezone? Is there any specific reason behind making them two separate fields? Can we not just use one variable for both?

felipe-augusto: It is not clear from the docker image [docs](https://github.com/halvves/mongodb-backup-s3):

```sh
TZ - timezone. default: US/Eastern

CRON_TZ - cron timezone. default: US/Eastern
```

I guess we can use the same variable

public/v2/apps/mongodb-backup-s3.json
felipe-augusto: done

githubsaturn:
Thanks!

PR #250, opened 2020-08-25
Add Kanboard SQLite version

seemsindie:
You can check it out running on http://kan.seemsindie.com/ (admin:admin)


public/v4/apps/kanboard-sqlite.yml
```diff
@@ -0,0 +1,31 @@
...
+          description: Check out their Docker page for the valid tags https://hub.docker.com/r/kanboard/kanboard/tags
+          validRegex: '/^([^\\s^\\/])+$/'
+        - id: $$cap_plugin_installer
+          label: Plugin installer
+          defaultValue: 'true'
+          description: Allow plugin installer
+          validRegex: '/^(true|false)$/'
+    instructions:
+        start: |-
+            Kanboard is a free and open source Kanban project management software.
```
githubsaturn, line 25: I suggest add a bit more info. Maybe a link to their website?

PR #251, opened 2020-08-25
Add Gogs (No DB) OCA

zuedev:
Here's a very simple Gogs OCA, with install still required. I don't know if this is appealing here due to the post-deployment install but I personally like that control and it's better than not having it there imo! 😉


githubsaturn:
> with install/customisation still present

Not sure if I completely understand this. Is it possible to create a docker compose that is working automatically, or is it the best gogs offers?

Also, are post installation steps in-line or user needs to read some documentation on how to finish the setup? If it's latter, there has to be proper instructions and additional notes referring to the document.

zuedev:
> Not sure if I completely understand this. Is it possible to create a docker compose that is working automatically, or is it the best gogs offers?

My bad, it was late! 😅 What I meant by this was that you can preconfigure a database connection for Gogs using environment variables, and I believe it may be best to keep this option open for those who want to choose a database driver to use as you'd have to go through installation anyway without a custom configuration file.

Unfortunately, the only [options](https://github.com/gogs/gogs/tree/master/docker#container-options) we have with Gogs are related to system-specific (and optional) tasks like backups.

There is an [issue](https://github.com/gogs/gogs/issues/3682#issuecomment-531090714) currently open about this on the Gogs repository.

> Also, are post installation steps in-line or user needs to read some documentation on how to finish the setup? If it's latter, there has to be proper instructions and additional notes referring to the document.

This may be subjective to me, but I think they are very easy to follow and are pretty self-explanatory. If this is deemed incorrect, I don't mind at all adding some more instructions.

githubsaturn:
> This may be subjective to me, but I think they are very easy to follow and are pretty self-explanatory. If this is deemed incorrect, I don't mind at all adding some more instructions.

Please add some level of documentation, and link to the page where it's explained on Gogs website perhaps.

> My bad, it was late! sweat_smile What I meant by this was that you can preconfigure a database connection for Gogs using environment variables, and I believe it may be best to keep this option open for those who want to choose a database driver to use as you'd have to go through installation anyway without a custom configuration file.

I am still unclear on this. Does Gogs need a database to function? If yes, then yes, we'll have to add the database to this one click app. Alternatively we can create two Gogs, similar to what we did with wordpress. We have a standard version and we have another variant without database which is aimed for more advanced users who need to set up their own database.

zuedev:
Unfortunately, it seems that it isn't trivial to preconfigure Gogs with just using the Dockerfile reference. I thought I could insert my own config via `dockerfileLines` and start it from there but I was mistaken. (See below for attempt)

The only choice going forward is to just have Gogs require an install step or to wait until they implement environment variables as Gitea has. I don't mind closing this PR if it doesn't seem too "OCA"-like for this repo. 😅

**Basic Idea (Doesn't Work)**
```!
services:
    $$cap_appname:
        restart: always
        volumes:
            - $$cap_appname-config:/data
        ports:
            - 10022:22
        caproverExtra:
            containerHttpPort: '3000'
            dockerfileLines:
                - FROM gogs/gogs:$$cap_gogs_version
                - RUN apk add --no-cache coreutils
                - RUN mkdir -p /data/gogs/conf/ && echo "<app.ini's base64>" | base64 -d > /data/gogs/conf/app.ini
                - ...
```

githubsaturn:
I don't mind merging this as long as we make it very clear in instructions.start and instructions.end about the steps that they'll have to take after the one click deploy is finished.

PR #253, opened 2020-08-25
Add BookStack OCA

zuedev:
Another one! This time with proper timezone configuration 😉 


public/v4/apps/bookstack.yml
```diff
@@ -0,0 +1,60 @@
...
+          description: Password to connect to the database
+          defaultValue: $$cap_gen_random_hex(16)
+          id: $$cap_DB_PASS
+        - label: Timezone
+          description: This is the timezone for the application, find yours at https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+          defaultValue: Europe/London
+          id: $$cap_timezone
+        - label: App URL
+          description: This is required for asset links to navigate the proxy
+          defaultValue: https://domain.tld
```
githubsaturn, line 46: You can change to `$$cap_appname.$$cap_root_domain`

public/v4/apps/bookstack.yml
```diff
@@ -0,0 +1,60 @@
...
+            PGID: 1000
+            PUID: 1000
+            TZ: $$cap_timezone
+            APP_URL: $$cap_appurl
+        image: linuxserver/bookstack:v0.29.3-ls99
+        restart: always
+        volumes:
+            - $$cap_appname-config:/config
+        caproverExtra:
+            containerHttpPort: '80'
```
githubsaturn, line 33: 80 is the default port, no need to define it explicitly

public/v4/apps/bookstack.yml
```diff
@@ -0,0 +1,60 @@
...
+            DB_HOST: srv-captain--$$cap_appname-mariadb
+            DB_USER: bookstack
+            DB_PASS: $$cap_DB_PASS
+            DB_DATABASE: bookstack
+            DB_PORT: 3306
+            PGID: 1000
+            PUID: 1000
+            TZ: $$cap_timezone
+            APP_URL: $$cap_appurl
+        image: linuxserver/bookstack:v0.29.3-ls99
```
githubsaturn, line 28: Generally it's a good idea to leave the version as a variable with a proper default value.

public/v4/apps/bookstack.yml
zuedev: 1d3a7f4

public/v4/apps/bookstack.yml
zuedev: Removed in 1d3a7f4

public/v4/apps/bookstack.yml
zuedev: Ah yes, I saw that in some of the other configs. With BookStack, there isn't really a major shift with the versions (to my knowledge) so I didn't think this was important. Should I still add the option for it? 🤔

githubsaturn: Yea I still think it's a good idea, otherwise the end user has no idea what version they are installing. Plus, when a newer version comes out, they can choose to change the newer tag.

zuedev: Added in 62a78bf

githubsaturn:
Thanks

PR #256, opened 2020-08-26
Passbolt - update default from email address env variable

rosinghal:
As per documentation in https://github.com/passbolt/passbolt_docker#configure-passbolt


githubsaturn:
Thanks!

PR #257, opened 2020-08-26
Add Cachet OCA

zuedev:
Would be cool to have an open-source status page in the repo! 😉 

https://cachethq.io/


public/v4/apps/cachet.yml
```diff
@@ -0,0 +1,54 @@
...
+            GitHub: https://github.com/CachetHQ/Cachet
+        end: |-
+            Aaaand you're done! 😄
+
+            Your service is available at http://$$cap_appname.$$cap_root_domain
+
+            If you need help with installation check out the docs! https://docs.cachethq.io/docs
+    displayName: Cachet
+    isOfficial: true
+    description: 📛 An open source status page system for everyone.
```
githubsaturn, line 53: Let's avoid using emoji in the description to be consistent with all other apps.

zuedev: Done in a23a2b1

PR #267, opened 2020-09-17
Latest drone, change from agent to docker runner

kerk1v:

- [yes] I have tested the template using the method described in README.md thoroughly
- [yes] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [yes] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [yes] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [no] Icon is added as a png file to the logos directory.

The existing one-click-app for Drone was using quite outdated versions of drone and the drone agent

I have updated to later Drone versions using the drone-agent

I have also updates the variables as needed for the newer versions. 

The oAuth2 secrets that need to be created in gitea can (and should) now be entered in the template form.

### DIFF
--- a/public/v4/apps/drone-gitea.yml
+++ b/public/v4/apps/drone-gitea.yml
@@ -1,14 +1,17 @@
 captainVersion: 4
 services:
     $$cap_appname-agent:
-        image: drone/agent:$$cap_drone_version
+        image: drone/agent:$$cap_drone_runner_version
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #269, opened 2020-09-21
Adding docker registry with a nifty UI

kerk1v:

- [☑️  ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- ☑️ [ ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [☑️  ] Icon is added as a png file to the logos directory.


public/v4/apps/docker-registry.yml
```diff
@@ -0,0 +1,40 @@
...
+caproverOneClickApp:
+    variables:
+        - id: $$cap_registry_version
+          label: Docker Registry image version
+          defaultValue: 2
+          description: Check out their Docker page for the valid tags https://hub.docker.com/_/registry?tab=tags - default is 2 as of 2020-09-20
+    instructions:
+        start: Standard docker registry with a Web UI for image maintenance.
+
+            Make sure you create a htpasswd file somewhere on your host with htpasswd -Bbn username password > /<path>/htpasswd.
```
githubsaturn, line 33: This is a bit complex. Can we, instead of a plain Dockerfile, we use dockerfileLines, and just create the password during container start up?

kerk1v: Let me investigate how this is done properly and I'll submit a new PR. Thanks for the suggestion.

githubsaturn: Could be something like:
```
FROM registry:$$cap_registry_version
RUN htpasswd -Bbn $$cap_username $$cap_password > /auth/htpasswd
```
Also, you'd have to remove `$$cap_appname-registry-auth:/auth` volume as it'll override it.

PootisPenserHere:
Is this feature still being worked on?

According to the docker documentation the work of authenticating users can be passed to a reverse proxy so there would be no need for a htpasswd file https://docs.docker.com/registry/recipes/nginx/  if there is a way to set up the basic auth from within the one click wizard that'd give users the option to launch their registries protected by caprover's basic auth or leave them open to the public.

PR #270, opened 2020-09-26
Allow use of symbols in the redis password

RodolfoSilva:
You can see the test case here: https://regex101.com/r/6ObzU2/2

image

image


githubsaturn:
You'd have to at least exclude `"` as it will breaks the JSON replacement when the template is being rendered.

RodolfoSilva:
@githubsaturn, done!

githubsaturn:
Thanks!

PR #273, opened 2020-10-03
Add Grav CMS

radex02:


radex02:
Why cant I pass the checks ? I tested it and it works perfectly

githubsaturn:
Run `npm run formatter-write`

radex02:
I've run `npm run formatter-write` and `npm run formatter` then `git push` but I stil fail the test here. How ?

githubsaturn:
You've probably not commit the changes imposed by formatter. This is what the validation does: https://github.com/caprover/one-click-apps/blob/master/.github/workflows/validate_apps.yml#L19

Regardless, I can merge and fix the formatter later on. But before that, why we go with an unofficial image rather than their official dockerfile: https://github.com/getgrav/docker-grav/blob/master/Dockerfile

radex02:
I use this image because the official one is not prebuilt

githubsaturn:
You'd still be able to include Dockerfile lines directly in one click app

radex02:
Would that be a best practice ?

githubsaturn:
If the build time is not insanely long, yes, that'd be ideal. But if it takes more than 1min or so to build, then unofficial prebuilt image is the way to go.

PR #274, opened 2020-10-03
[FIX] fix persistent storage for vscode server to prevent ssh and plu…

LiquidITGuy:
Few important folder are missing to prevent lost on reboot.
First one is `/root/.ssh` (we are logged as root so if we reboot, we lost our ssh private key)
Second one is `/root/.local/share/code-server` who host all the plugins and themes informations

- [ ] I have tested the template using the method described in README.md thoroughly => didn't do it, just do this on my deployed app. Will try it when got more time

### DIFF
--- a/public/v2/apps/vscode-coder.json
+++ b/public/v2/apps/vscode-coder.json
@@ -19,7 +19,9 @@
             "restart":"always",
             "volumes":[
                "$$cap_appname-config-data:/home/coder/.local/share/code-server",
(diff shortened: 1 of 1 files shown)


githubsaturn:
`v2` is no longer in the repo. The new schema directory is `v4`. Feel free to edit here:
https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/vscode-coder.yml

PR #278, opened 2020-10-08
Do not force cap_appname-gitea as a sub

AkgunFatih:
let users decide the app name theirself
and don't force it as **cap_appname-gitea.domain.com**


githubsaturn:
Thanks

PR #285, opened 2020-10-22
Adding organizr logo and captain file

Scarzer:

I've added the Organizr dashboard for quick deployment. The only issue I'd like to point to is the use of the `amd64` tag. There really isn't too many good choices provided by Organizr (yet)


githubsaturn:
Thanks!

PR #292, opened 2020-10-30
Adding Plausible Analytics.

thomascenni:

- [X] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version (only for service mail with image: bytemark/smtp I am using latest tag because it's the only one)


public/v4/apps/plausible.yml
```diff
@@ -0,0 +1,131 @@
...
+            SECRET_KEY_BASE: $$cap_gen_random_hex(64)
+            SIGNING_SALT: $$cap_gen_random_hex(24)
+            DISABLE_REGISTRATION: "true"
+            DISABLE_SUBSCRIPTION: "true"
+            MAILER_EMAIL: $$cap_RELAY_USERNAME
+            SMTP_HOST_ADDR: srv-captain--$$cap_appname-mail
+            SMTP_HOST_PORT: "25"
+            GEOLITE2_COUNTRY_DB: "/geoip/GeoLite2-Country.mmdb"
+        volumes:
+            - $$cap_appname-geoip-data:/geoip:ro
```
githubsaturn, line 68: drop `:ro` as it's not respected by CapRover

githubsaturn (commented):
Thanks for submitting this, looks like a large application. Would you be able to add some more details to instructions.start?
- The minimum/recommended RAM and CPU
- Also, list out all the different images that is being installed since it's a collection of various images maintained by different parties.

public/v4/apps/plausible.yml
thomascenni: In their page:
https://docs.plausible.io/self-hosting
There are no minimum requirements.
I'll list the different images used in the description.

githubsaturn:
Thanks!

LittleBigFrog:
Hello, 

Unfamiliar with clickhouse. Hope it is the right place to ask.

Plausible on a 5$ digitalocean droplet uses 100% of CPU.
It seems to be linked to Clickhouse, thanks to investigation by @RealOrangeOne :

https://theorangeone.net/posts/calming-down-clickhouse/
Could we have clickhouse low ressource setup with default no logging via parameters?

Thank you

PR #293, opened 2020-10-30
Add one-click app for keycloak-only

Andrewpk:


public/v4/apps/keycloak-only.yml
```diff
@@ -0,0 +1,64 @@
...
+             For documentation, see: https://www.keycloak.org/documentation.html
+
+            For source code, see: https://github.com/keycloak/keycloak
+
+
+            For docker image, see: https://hub.docker.com/r/jboss/keycloak
+
+              Enter your Keycloak Configuration parameters and click on next. It will take about a minute for the process to finish.
+        end: Keycloak is deployed and available as srv-captain--$$cap_appname:8080 to other apps
+    displayName: ''
```
githubsaturn, line 61: Can you update this to "Keycloak - No Database" to make it more clear for users?

PR #294, opened 2020-11-01
Wordpress Databases Password Fix

touhidshaikh:
Allow special symbols in WordPress's databases password field.

the current regex is not allowing the user to put the special character such as @, $  etc.

**current regex status:** https://regex101.com/r/AQHqeO/1 
**modified or fixed regex:** https://regex101.com/r/fwFYSJ/1

- [x] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end-users to get started.

### DIFF
--- a/public/v4/apps/wordpress.yml
+++ b/public/v4/apps/wordpress.yml
@@ -32,7 +32,7 @@ caproverOneClickApp:
         - id: $$cap_db_pass
           label: Database password
           description: ''
-          validRegex: /^([a-zA-Z0-9])+$/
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks for the suggestion! Could you also negate single quote `'` ?

touhidshaikh:
> Thanks for the suggestion! Could you also negate single quote `'` and backslash `\` ?
@githubsaturn thanks. the update is done. please review and let me know.

githubsaturn:
Thanks!

PR #296, opened 2020-11-03
Added InvoicePlane.

thomascenni:

- [X] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started. There are default data and randomly generated data.
- [X] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version. Yes, except from the image woahbase/alpine-mysql that has only the latest tag.


public/v4/apps/invoiceplane.yml
```diff
@@ -0,0 +1,67 @@
+captainVersion: 4
+
+services:
+    $$cap_appname-db:
+        image: woahbase/alpine-mysql:x86_64
```
githubsaturn, line 5: any reason not to use the official mysql image?

githubsaturn (commented):
Thanks for adding this!

Looks like there is some conflict with your previous PR where you added plausible.

public/v4/apps/invoiceplane.yml
thomascenni: The official MySQL image is 145 MB, alpine-mysql is 70 MB. 
But we can use teh official one if you prefer.
Concerning the conflicts, I do not understand, because I use prettier and with prettier the string are enclosed in double quotes (") automatically when I format the YML, not with single quotes (').

githubsaturn: > The official MySQL image is 145 MB, alpine-mysql is 70 MB.
> But we can use the official one if you prefer.
Yea I think it's better to go with the official version as this version doesn't even allow the users to set the version of MySQL engine. The tags only include the architecture. 

> Concerning the conflicts, I do not understand, because I use prettier and with prettier the string are enclosed in double quotes (") automatically when I format the YML, not with single quotes (').

You just need to back merge master into your branch, or you can rebase. Both work.
If it complicates things for you, you can go with the ultimate option:
- Back up your template
- Delete your forked repository
- Fork again and create the PR

thomascenni: And I have just discovered that there is the option "--single-quote" in prettier that I was not using !
I'll try to resolve teh conflicts. Thanks.

githubsaturn: As for prettier, you can just simply run
```
npm i
npm run formatter-write
```

githubsaturn:
Thanks!

Issue #15, opened 2019-02-02
Suggetion: searx


githubsaturn:
https://github.com/asciimoo/searx

ernaneluis:
I started here the PR, any help would be great https://github.com/caprover/one-click-apps/pull/192

PR #309, opened 2020-12-03
Added Searx

kgnfth:
#15
#192


public/v4/apps/searx.yml
```diff
@@ -0,0 +1,96 @@
...
+        - label: Path to Filtron rules.json file
+          defaultValue: /var/lib/docker/volumes/captain--searx-filtron-rules/_data/rules.json
+          id: $$cap_FILTRON_RULES
+    instructions:
+        start: >-
+            Privacy-respecting, hackable metasearch engine.
+
+            Mayday! Mayday! Mayday!
+
+            First of all put the content of https://github.com/searx/searx-docker/blob/master/rules.json
```
githubsaturn, line 87: If it's literally just this file, why not use this line in Dockerfile?
```
ADD https://github.com/searx/searx-docker/blob/master/rules.json /etc/filtron/rules.json
```

Then we don't even need to cross map the volumes

public/v4/apps/searx.yml
```diff
@@ -0,0 +1,96 @@
...
+          label: Morty Docker Version Hash
+          defaultValue: 'sha256:cb930c25254337e7a3abaf6daab326c663a4cdf1cbf4ae040ff7f670493df7fd'
+          description: Starts with 'sha256:'.  Find on https://hub.docker.com/r/dalf/morty/tags/
+        - id: $$cap_checker_version
+          label: Searx-checker Docker Version Hash
+          defaultValue: 'sha256:8ebda54ed8854e1300a4670e2bba86a5d2f4ceeb5c39290e4cc20df22bf974be'
+          description: Starts with 'sha256:'.  Find on https://hub.docker.com/r/searx/searx-checker/tags/
+        - label: Searx base url
+          description: Base url to your Searx
+          defaultValue: https://searx.app.gocloud.fun
```
githubsaturn, line 68: If this is the URL from the app itself, it doesn't need to be a variable. 

You can just hardcode this as the env var:
```
http://$$cap_appname.$$cap_root_domain
```

public/v4/apps/searx.yml
```diff
@@ -0,0 +1,96 @@
...
+          id: $$cap_SEARX_BASEURL
+        - label: Morty URL
+          description: Base url to your Morty
+          defaultValue: https://searx-morty.app.gocloud.fun
```
githubsaturn, line 72: Not sure what "Morty" is. But this default value seems be just an example, right? If that's the case, move it to description please.

The idea behind default values is that if the user doesn't make any changes to them, they just work.

kgnfth: See https://github.com/dalf/morty 

The main goal of morty is to provide a result proxy for searx

Features:
- HTML sanitization
- Rewrites HTML/CSS external references to locals
- JavaScript blocking
- No Cookies forwarded
- No Referrers
- No Caching/Etag
- Supports GET/POST forms and IFrames
- Optional HMAC URL verifier key to prevent service abuse

example

https://searx-morty.app.gocloud.fun/?mortyurl=https%3A%2F%2Fmedium.com%2Fswlh%2Fcaprover-the-definitive-guide-90076405aae4&mortyhash=dde3bfae15fc28782232c2cde20e9cc5697fa8b182279fdcad7ff8d7ff3d9437

public/v4/apps/searx.yml
kgnfth: Learned something new!
Thx

Issue #288, opened 2020-10-25
App Suggestion: Monica


doebi:
**Please tell us what app you'd like to see on CapRover as a one-click app.**

https://github.com/monicahq/monica

**Do you know if there is any official Docker image for the app?**

Their recommended setup is based on docker: https://github.com/monicahq/monica/blob/master/docs/installation/providers/docker.md

kgnfth:
@githubsaturn  Can be closed

PR #311, opened 2020-12-05
Added Monica #288

kgnfth:


public/v4/apps/monica.yml
```diff
@@ -0,0 +1,60 @@
+captainVersion: 4
+services:
+    # Monica
+    $$cap_appname:
+        depends_on:
+            - $$cap_appname-mariadb
+        image: monica:apache
```
githubsaturn, line 7: I am assuming they don't have versioned tags, right?

If that's the case, I suggest we use sha hash to ensure that the one-click app won't break if they change the image and change the requirement in the future.

public/v4/apps/monica.yml
```diff
@@ -0,0 +1,60 @@
...
+        environment:
+            APP_KEY: $$cap_gen_random_hex(32)
+            DB_HOST: srv-captain--$$cap_appname-mariadb
+            DB_DATABASE: $$cap_mariadb-db
+            DB_USERNAME: $$cap_mariadb-user
+            DB_PASSWORD: $$cap_mariadb-pass
+        volumes:
+            - $$cap_appname-storage:/var/www/html/storage
+        caproverExtra:
+            containerHttpPort: '80'
```
githubsaturn, line 18: minor: port 80 is the default and not needed.

githubsaturn:
Thanks!

aniketpanjwani:
I'm on CapRover 1.8.2, and this one-click app isn't working for me. After deployment, I'm getting an NGINX 502 Error page with a link to [these docs](https://caprover.com/docs/troubleshooting.html#successful-deploy-but-502-bad-gateway-error). I tried deploying another one-click app, CodiMD, to make sure the problem isn't with all one-click apps, and that app works fine. The only thing I modified in the one-click app configs for Monica was the name of the app, which was left blank, so I called it "monica".

Don't have time for a full trouble-shooting now, but I thought I'd make a quick comment in case anyone else is experiencing similar issues.

aniketpanjwani:
I think I figured out the issue - I was leaving the MariaDB password blank, but that's required. Got it working now!

kgnfth:
Works fine here http://monica.app.gocloud.fun

kgnfth:
Works fine here https://monica.app.gocloud.fun

to fix your styles you have to add `APP_ENV=production` in your app env config

See https://hub.docker.com/_/monica?tab=description and scroll down all the way to `Make Monica available from the internet`

aniketpanjwani:
@kgnfth Thanks that fixed it for me!

githubsaturn:
@kgnfth  - Thanks for pointing out `APP_ENV=production` 

Is there any downside to including this in the default config?

kgnfth:
@githubsaturn No, but i think i missed this option when i first created this template
i will send a pull now

PR #319, opened 2020-12-15
Wordpress nginx only

SwapRane12:


githubsaturn:
Thanks for the PR. Can you please outline the point of this variant as opposed to [WordPress Only](https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/wordpress-only.yml)?

SwapRane12:
> Thanks for the PR. Can you please outline the point of this variant as opposed to [WordPress Only](https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/wordpress-only.yml)? 
> 

That means, without database, this is for those who want to use wordpress with php fpm and nginx,

githubsaturn:
I see. Okay then, can you please update the name, and instruction.start , looks like they are copied from WordPress Only file and they do not reflect php fpm etc. Also please set isOfficial to false as this one click app is using a third party nginx image.

https://github.com/caprover/one-click-apps/blob/af172b6680583487bdeacf230d7abaf9b57f4811/public/v4/apps/wordpress-only.yml#L40-L54

SwapRane12:
> I see. Okay then, can you please update the **displayName**, and **instruction.start** , looks like they are copied from WordPress Only file and they do not reflect php fpm etc. Also please set **isOfficial** to false as this one click app is using a third party nginx image.
> 
> https://github.com/caprover/one-click-apps/blob/af172b6680583487bdeacf230d7abaf9b57f4811/public/v4/apps/wordpress-only.yml#L40-L54

Yes they are coppied,  I will update it when I got access to my computer.

holygekko:
@SwapRane12 I was looking for something like this, but can't get it to work.
I followed all instructions:
- I have an existing mysql container (from the default wp one click app)
- created a new db within it with a new user that can acces it
- filled in all variables as requested.

But the nginx container keep giving this error: `[error] 15#15: *1 connect() failed (111: Connection refused) while connecting to upstream, client: 10.0.0.10, server: _, request: "GET / HTTP/1.0", upstream: "fastcgi://10.0.0.19:9000", host: "lempwp-nginx.xxx.xxx.com"`

The wordpress container gives me this: `AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 10.0.0.33. Set the 'ServerName' directive globally to suppress this message`

So it seems nginx is trying to reach wordpress on 10.0.0.19 instead of 10.0.0.33?

Can you advise me what to do?

Issue #317, opened 2020-12-15
App Suggestion: wallabag


fi0:
https://hub.docker.com/r/wallabag/wallabag

PR #320, opened 2020-12-15
added Wallabag #317

kgnfth:
Demo https://wall.app.gocloud.fun
user:`wallabag`
pass:`wallabag`

Demo will remain until merged


githubsaturn:
Thanks!

pronoiac:
It might be worth noting that Wallabag uses email to activate accounts, and so having email setup is a dependency. Should we note this in the description? 

> Originally posted as #484.

kgnfth:
> It might be worth noting that Wallabag uses email to activate accounts, and so having email setup is a dependency. Should we note this in the description?
> 
> > Originally posted as #484.

I created this setup as a single user.
Thats the reason i did not include the `SYMFONY__ENV__MAILER_*` variables in this setup, not everyone uses a email service.
If u don't have a email server, you can always activate the user manually from the wallabag dashboard.

pronoiac:
Ah, I realized the user/pass from the demo above also works for the installed app. Thanks!

kgnfth:
> Ah, I realized the user/pass from the demo above also works for the installed app. Thanks!

Yes, at the end of the installation is it also displayed. See https://github.com/caprover/one-click-apps/blob/938fe9ff2c517f8af9ed9fc3a97adb3540c89991/public/v4/apps/wallabag.yml#L66

PR #321, opened 2020-12-17
Form.io

aqidd:


public/v4/apps/formio.yml
```diff
@@ -0,0 +1,94 @@
...
+          label: SMTP port
+          defaultValue: '587'
+          description: The port for your smtp server
+        - id: $$cap_smtp_user
+          label: SMTP user
+          defaultValue: your_email@gmail.com
+          description: The user for your smtp server
+        - id: $$cap_smtp_pass
+          label: SMTP password
+          defaultValue: sTr0nG_p@s$wo%d
```
githubsaturn, line 83: Use hex generator, example:
https://github.com/caprover/one-click-apps/blob/a37430b15cec4ba4bceab3d434f608536fe973b8/public/v4/apps/whoopsmonitor.yml#L143

public/v4/apps/formio.yml
```diff
@@ -0,0 +1,94 @@
...
+          label: SMPT host
+          defaultValue: smtp.gmail.com
+          description: The host for your smtp server
+        - id: $$cap_smtp_port
+          label: SMTP port
+          defaultValue: '587'
+          description: The port for your smtp server
+        - id: $$cap_smtp_user
+          label: SMTP user
+          defaultValue: your_email@gmail.com
```
githubsaturn, line 79: Remove the default value to indicate that the user needs to modify this, maybe move it to label.

Same with other default values that are not actually default values, and they are just examples. Let's move them to "label"

githubsaturn (commented):
Thanks! Just some minor comments.

public/v4/apps/formio.yml
aqidd: Thanks! I decided to remove all the default values for the SMTP as they already have a label

githubsaturn:
Thanks

PR #325, opened 2020-12-19
Weblate translation management service

sergioisidoro:
Weblate is a translation management service. 

~~'m still testing this~~ Everything seems to be working ok :)


public/v4/apps/weblate.yml
```diff
@@ -0,0 +1,88 @@
...
+          description: (eg weblate@example.com)
+        - id: $$cap_open_registration
+          label: Allow open registration of people (0/1)
+          defaultValue: 0
+          description: Do you allow people to register openly to the platform?
+    instructions:
+        start: Open-source collaborative localization tools
+        end: weblate is deployed and available as srv-captain--$$cap_appname:80.  Note that the application may take up to ten minutes to become available.
+    displayName: weblate
+    isOfficial: false
```
kgnfth, line 86: weblate/weblate is official image so you can set
isOfficial: false to true

sergioisidoro: Oh! I thought official meant that this was something that this was some sort of official apps maintained by the CapRover project (as in, not the community). Thanks I'll fix this :)

sergioisidoro:
I found a bug in the app volume that had pretty nasty consequences of losing data. 

I will let this service run for a little longer in my cluster to make sure I didn't overlook anything.

public/v4/apps/weblate.yml
```diff
@@ -0,0 +1,88 @@
...
+            - $$cap_appname-postgres-data:/var/lib/postgresql/data
+        restart: always
+        environment:
+            POSTGRES_USER: weblate
+            POSTGRES_PASSWORD: $$cap_pg_pass
+            POSTGRES_DB: weblate
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+    $$cap_appname:
+        image: weblate/weblate
```
kgnfth, line 21: you forgot $$cap_weblate_version

public/v4/apps/weblate.yml
```diff
@@ -0,0 +1,88 @@
...
+        - id: $$cap_postgres_version
+          label: Postgres Docker version
+          defaultValue: 13-alpine
+          description: Check out their Docker page for the valid versions https://hub.docker.com/r/library/postgres
+        - id: $$cap_redis_version
+          label: Redis Docker version
+          defaultValue: 6-alpine
+          description: Check out their Docker page for the valid versions https://hub.docker.com/r/library/redis
+        - id: $$cap_weblate_version
+          label: weblate Docker Image SHA Hash
```
kgnfth, line 59: Based on the label
if we need to use hash instead of tag
you need use image: `weblate/weblate@$$cap_weblate_version` at line 21

githubsaturn: Generally, if a proper tag is available, it should be the preferred option as it's much more readable than the hash. The only benefit of hash is that it's not mutable.

sergioisidoro:
Hey there. 

Thanks for your help and comments. I think all the comments you mentioned have been addresses. The mistake on the volume mount is also now corrected (the path I was mounted was incorrect)

Let me know if there's anything else I can do, and sorry for the delay in addressing this :)

githubsaturn:
Looks good. Can you please replace latest with a concrete version?

githubsaturn:
Awesome, thanks!

Issue #328, opened 2020-12-22
App Suggestion: [AdGuard Home]


f97:
AdGuard Home is a network-wide software for blocking ads & tracking. After you set it up, it’ll cover ALL your home devices, and you don’t need any client-side software for that. With the rise of Internet-Of-Things and connected devices, it becomes more and more important to be able to control your whole network.

---------
Docker image : [adguard/adguardhome](https://hub.docker.com/r/adguard/adguardhome)

kgnfth:
You can try this

```yaml
captainVersion: 4
services:
    # AdGuard
    $$cap_appname:
        image: adguard/adguardhome:$$cap_adguard_version
        restart: always
        ports:
            - 53:53
            - 67:67
            - 68:68
            - 8081:80
            - 4433:443
            - 853:853
        volumes:
            - $$cap_appname-work:/opt/adguardhome/work
            - $$cap_appname-conf:/opt/adguardhome/conf
        caproverExtra:
            containerHttpPort: '3000'
caproverOneClickApp:
    variables:
        - id: $$cap_adguard_version
          label: AdGuard Home Version
          defaultValue: 'v0.104.3'
          description: Checkout their docker page for the valid tags https://hub.docker.com/r/adguard/adguardhome/tags
    instructions:
        start: >-
            AdGuard Home is a fast and easy way to block ads on any device and protect kids online!
        end: >-
            Aaaand you're done! 😄
            Your service is available at http://$$cap_appname.$$cap_root_domain
    displayName: 'AdGuard Home'
    isOfficial: true
    description: AdGuard Home is a fast and easy way to block ads on any device and protect kids online!
    documentation: Taken from https://hub.docker.com/r/adguard/adguardhome
```

after deployed go to the app settings and enable https and force https
after that go to your url and on the installation steps change port 80 to 3000
when u are done you will be redirected to yourul.com:3000
just remove the the port from the url and you can access the login page

DHCP wont work because we cant change the network host for the docker using the template

adblocking works
all you have to do next is
after you configure your filters
you can use your host ip to block services or ads

i can create a pull request if you like it

Pls let me know

marcusique:
@kgnfth thanks for the template. 

Do you by any chance know how to make AdGuard identify clients that use AdGuard DNS? As of now I have only one client in the list and it’s 10.0.0.2. I assume it’s somehow related to the docket-bridged network instead of host.

kgnfth:
> @kgnfth thanks for the template.
> 
[...]

PR #330, opened 2020-12-22
:sparkles: Added AdGuard Home

kgnfth:
#328


public/v4/apps/adguard-home.yml
```diff
@@ -0,0 +1,42 @@
...
+    # AdGuard
+    $$cap_appname:
+        image: adguard/adguardhome:$$cap_adguard_version
+        restart: always
+        ports:
+            - 53:53
+            - 67:67
+            - 68:68
+            - 8081:80
+            - 4433:443
```
githubsaturn, line 12: Are these all required? Wouldn't nginx regular proxy be enough for 80/443?

kgnfth: yea let me fix that

githubsaturn:
Thanks!

PR #334, opened 2020-12-24
✨ Added Duplicati & removed port mapping in lychee 🔧

kgnfth:

Fully Tested

image


githubsaturn:
This is really nice! Thanks!

githubsaturn:
> I tried adding /var/lib/docker/volumes as a persistent source but caprover would not accept it.

What do you mean CapRover wouldn't accept it? You can map any directory of the host to any directory in the container.

coffseducation:
I could not add more than one
/source /directory
Caprover threw an error message.

I left the source at / to access all directories.
Duplicati give a guid and puid as 1000 in settings.
I changed this to root  "0" and can now access the docker volumes, so in the end did not need multiple sources configured.

nb. Should the Duplicati One Click app uid settings be changed to "0" - or at least have a note about this in setup?
Reason: Noobies like myself will choose a One Click app, and go with defaults that have data in Docker volumes rather than in user directories. 0 would make more sense and save the new user from spending days finding the answer.

Thanks.

githubsaturn:
> I could not add more than one
/source /directory
Caprover threw an error message.

What was the error message?

Log in to https://captain.server.demo.caprover.com with password of `captain42` and check out `test-directory-mapping`

image

coffseducation:
image
"Something bad happened"

Two sources probably confuses Duplicati.
Is there somewhere I can find logs?

kgnfth:
> image "Something bad happened"
> 
> Two sources probably confuses Duplicati. Is there somewhere I can find logs?

That's because `/var/lib/docker/` is already in the root folder `/`
Just remove the source `/var/lib/docker/`

githubsaturn:
@coffseducation - You can't map two different directories to the same directory in the container. Mappings should be 1:1

PR #336, opened 2020-12-29
✨ Added Statping & Radarr

kgnfth:


public/v4/apps/radarr.yml
```diff
@@ -0,0 +1,66 @@
...
+            - $$cap_appname-config:/config
+            - $$cap_path_to_movies:/movies
+            - $$cap_path_to_downloads:/downloads
+        caproverExtra:
+            containerHttpPort: 7878
+caproverOneClickApp:
+    variables:
+        - id: $$cap_radarr_version
+          label: Radarr Version
+          defaultValue: amd64-latest
```
githubsaturn, line 20: Let's change this to a fixed version.

Also, looks like they have multiarch tags as well. Let's use those:
image

kgnfth: those multi arch is equivalent to [version-3.0.1.4259](https://hub.docker.com/layers/linuxserver/radarr/version-3.0.1.4259/images/sha256-9e65bfed06f9cadeef2b04ec5fd42f08316c03858ade90e2eb407d3176a106df?context=explore)
should i change it to that ?

githubsaturn: Yea - that looks good

kgnfth: Oki

githubsaturn: FYI - since you're contributing a lot to these one click apps it's good for you to know the rationale behind these guidelines, e.g., not using `latest` tag. The reason behind that is that a new version might have a different directory mapping, different env vars and etc. Using a specific tag ensures that the one click app will continue to work forever.

githubsaturn:
Awesome!

public/v4/apps/radarr.yml
kgnfth: Yea i forgot about that

chocovish (commented):
DB_HOST env variable is not working. even after changing the host env variable it is still trying to connect with the srv_captain__statping-postgres (and failing, as i renamed the statping-postgres service)

Issue #200, opened 2020-06-27
App Suggestion: Mastodon


rymdluo:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
Twiiter alternative Mastodon: https://github.com/tootsuite/mastodon

**Do you know if there is any official Docker image for the app?**
Dockerfile: https://github.com/tootsuite/mastodon/blob/master/Dockerfile
Compose: https://github.com/tootsuite/mastodon/blob/master/docker-compose.yml

lisaiceland:
How do I add an admin user?

somnathrakshit:
I am having troubles sending confirmation email while creating an account from the web UI

nghialele:
> 

Please provide more details in what you mean "troubles". 
Or at least put in a copy of the log where the error was logged.

I suppose you are having issues with SMTP mail server. Are you managed to solve it?

maietta:
This is an ongoing issue as of today, Dec 22.

This must be a configuration problem and the provided One-Click-App needs further tweaking to be correct.

PR #338, opened 2020-12-31
✨ Added Mastodon

kgnfth:
Demo: https://mastodon.app.gocloud.fun

Edit: i got permission issue when trying to upload a profile picture in this mastodon instance
but i fixed it by running `chown -R 991:991 /var/lib/docker/volumes/captain--mastodon-system`
https://github.com/tootsuite/documentation/pull/596/files

Should i document this in this template ?
and if by changing the default user folder owner will this affect caprover ?


kgnfth:
#200

githubsaturn:
For the `chown` command, can you try augmenting the CMD line, i.e.:
```
CMD rm -f /mastodon/tmp/pids/server.pid; bundle exec rails db:setup; chown -R 991:991 /mastodon/public/system; bundle exec rails s -b 0.0.0.0 -p 3000
```

kgnfth:
> For the `chown` command, can you try augmenting the CMD line, i.e.:
> 
> ```
> CMD rm -f /mastodon/tmp/pids/server.pid; bundle exec rails db:setup; chown -R 991:991 /mastodon/public/system; bundle exec rails s -b 0.0.0.0 -p 3000
> ```

I will try now

githubsaturn:
> > For the `chown` command, can you try augmenting the CMD line, i.e.:
> > ```
> > CMD rm -f /mastodon/tmp/pids/server.pid; bundle exec rails db:setup; chown -R 991:991 /mastodon/public/system; bundle exec rails s -b 0.0.0.0 -p 3000
> > ```
> 
> I will try now
> @githubsaturn
> Just tested it but it does not work
> Also running the command inside the container gives the following error
> `chown: changing ownership of 'public/system': Operation not permitted`

Looking at their dockerfile, looks like the [USER is not root](https://github.com/tootsuite/mastodon/blob/master/Dockerfile#L134). So that's why it doesn't work. 

Alternatively, you can just point out this quirk in instructions.

kgnfth:
> > > For the `chown` command, can you try augmenting the CMD line, i.e.:
> > > ```
> > > CMD rm -f /mastodon/tmp/pids/server.pid; bundle exec rails db:setup; chown -R 991:991 /mastodon/public/system; bundle exec rails s -b 0.0.0.0 -p 3000
> > > ```
> > 
> > 
> > I will try now
> > @githubsaturn
> > Just tested it but it does not work
> > Also running the command inside the container gives the following error
> > `chown: changing ownership of 'public/system': Operation not permitted`
> 
> Looking at their dockerfile, looks like the [USER is not root](https://github.com/tootsuite/mastodon/blob/master/Dockerfile#L134). So that's why it doesn't work.
> 
> Alternatively, you can just point out this quirk in instructions.

Done

its 12pm here
🎇🎉🎇🎉 Happy New Year 🎇🎉🎇🎉

githubsaturn:
Thanks, and Happy New Years!

somnathrakshit:
Can you please add an example input for these variables? It is hard to understand what kind of input those fields require?
https://github.com/caprover/one-click-apps/blob/5006081dc44e3c610c27f739c55525eb942ac2cf/public/v4/apps/mastodon.yml#L187

kgnfth:
> Can you please add an example input for these variables? It is hard to understand what kind of input those fields require?
> 
> https://github.com/caprover/one-click-apps/blob/5006081dc44e3c610c27f739c55525eb942ac2cf/public/v4/apps/mastodon.yml#L187

If your mail server configuration requires tls, set to `true `otherwise to `false`

PR #346, opened 2021-01-14
Add Shiori, a simple link saving service

nathanielknight:
This commit adds a configuration file and logo for [Shiori], a simple
link saving service. Shiori can also save copies of the pages you save
with it.

[Shiori]: https://github.com/go-shiori/shiori


public/v4/apps/shiori.yml
```diff
@@ -0,0 +1,31 @@
...
+      containerHttpPort: '8080'
+
+caproverOneClickApp:
+  isOfficial: false
+  description: "A simple bookmark manager built with Go."
+  instructions:
+    start: >
+      This app stores its data in a SQLite database in a Docker volume.
+
+      Once deployed, this app will be accessible with default credentials. If
```
githubsaturn, line 21: Let's call out what the default creds are:
```
username: shiori
password: gopher
```

githubsaturn (commented):
Thanks, just one minor comment.

public/v4/apps/shiori.yml
nathanielknight: I've updated the instructions so that the username and password are in both the "start" and "end" pages.

Thanks for the feedback 🙂

githubsaturn:
Thanks!

PR #364, opened 2021-02-20
Use $$cap_root_domain to help facilitate the easy setup of sentry

andrewgy8:
According to this [comment](https://github.com/caprover/one-click-apps/pull/60#issuecomment-488957376) and this [comment](https://github.com/caprover/one-click-apps/pull/60#issuecomment-521431892), it is now possible to ease the installation of sentry.

### DIFF
--- a/public/v4/apps/sentry.yml
+++ b/public/v4/apps/sentry.yml
@@ -91,7 +91,7 @@ services:
                 - "RUN echo 'mail.use-tls: true' >> /etc/sentry/config.yml"
                 - 'RUN echo ''mail.username: ""'' >> /etc/sentry/config.yml'
                 - 'RUN echo ''system.admin-email: ""'' >> /etc/sentry/config.yml'
-                - 'RUN echo ''system.url-prefix: "$$cap_root_domain"'' >> /etc/sentry/config.yml'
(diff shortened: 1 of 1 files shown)


githubsaturn:
cap_root_domain is the root domain. You need to prefix it with the app name

andrewgy8:
Ah right! It would be something like this? `$$cap_appname-sentry.$$cap_root_domain`

andrewgy8:
Can confirm that it works as expected 
Screen Shot 2021-02-20 at 6 59 10 PM

githubsaturn:
Thanks!

public/v4/apps/sentry.yml
```diff
@@ -91,7 +91,7 @@ services:
                 - "RUN echo 'mail.use-tls: true' >> /etc/sentry/config.yml"
                 - 'RUN echo ''mail.username: ""'' >> /etc/sentry/config.yml'
                 - 'RUN echo ''system.admin-email: ""'' >> /etc/sentry/config.yml'
-                - 'RUN echo ''system.url-prefix: ""'' >> /etc/sentry/config.yml'
+                - 'RUN echo ''system.url-prefix: "http://$$cap_appname-sentry.$$cap_root_domain"'' >> /etc/sentry/config.yml'
```
alianwar, line 94: You're assuming that your sentry will run on `http`. After installation, there is no way to switch to `https` or update `system.url-prefix` if you connected a custom domain to your caprover app.

PR #365, opened 2021-02-20
Remove the suffix of "sentry" from the url prefix

andrewgy8:
Ooops, from #364 I accidentally added the 'sentry' suffix to the app name. Therefore the DSN comes out pointing to the wrong location. I found this out on a recent deploy I just did with the changes here 🤦 .


githubsaturn:
Thanks for the quick fix!

PR #366, opened 2021-02-23
Fix: Chatwoot setup command

sojan-official:


sojan-official:
ref: https://github.com/chatwoot/chatwoot/issues/1744

githubsaturn:
Thank you @sojan-official for the fix.

Quick question though, did it break for the newer versions or the embedded version in CapRover (1.9) too?

We avoid using the `latest` tag to avoid these breaking changes, but this shouldn't have broken for a version that was already published, right? I am curious what happened.

sojan-official:
@githubsaturn thanks a lot for getting this merged.
It didn't break for the `1.9` embedded version. 

But In our documentation, we ask users to replace the version value with our latest tag. Hence we started getting support requests :)

PR #369, opened 2021-03-11
Tidy instructions spelling & spacing

edwardch:
Updated spacing, layout, spelling, etc in display names and start/end instructions


edwardch:
>  Checking formatting...
> public/v4/apps/collabora-online.yml
> public/v4/apps/squidex.yml
> public/v4/apps/the-lounge.yml
> Code style issues found in the above file(s). Forgot to run Prettier?

Seems I did something wrong with running my Prettier (Windows). 

Will look again

githubsaturn:
Thanks for the fixes! Just be careful, this is going to cause conflicts very soon if we make any changes, so I suggest you fix the formatter and update the PR soon.

edwardch:
The validate_apps didn't pick up errors on macOS either, and all I can think of is that my node/npm versions are different to the one in setup-node build action. I had to find these the good old-fashioned way - by eye.

I'm glad the check is in place!

githubsaturn:
Thanks!

Was it automated in some way? If so, feel free to move the logic to the repo as a github action so that we won't break these again.

edwardch:
Sure thing!

Unfortunately nothing automated from my side. I'll try to explore some ideas though!

Thanks for accepting the PR.

PR #371, opened 2021-03-14
Bsr normalize descriptions

LiquidITGuy:
Hi,
during my last Pull request I found a lot of checkout (used as check out)
I fixed all the "check out/checkout" by putting check out (and adding a uppercase when it was at the begin of the line)


LiquidITGuy:
edit : may be a lot of duplicated thing with the PR 369 (who has done more check than me :) )

githubsaturn:
yes, #369 seems to have more changes. I'll merge that, and if there is anything else left, feel free to open another PR.

PR #384, opened 2021-04-12
Modification env variable for N8N database usage

laruche:
according to : https://github.com/n8n-io/n8n/blob/master/docker/compose/withPostgres/docker-compose.yml


githubsaturn:
Since which version the env vars changed? Do we need to update the default version too?

laruche:
> Since which version the env vars changed? Do we need to update the default version too?

Is not explicit in their documentation
but, this example file was created in 08/2019
and default version 0.67.3 have only one year... so it's working :)

githubsaturn:
Thanks!

PR #397, opened 2021-04-20
Upgrade bitwarden_rs from 1.15.1 to 1.20.0

LiquidITGuy:

### DIFF
--- a/public/v4/apps/strapi.yml
+++ b/public/v4/apps/strapi.yml
@@ -33,7 +33,7 @@ caproverOneClickApp:
     variables:
         - id: $$cap_strapi_version
           label: Strapi Version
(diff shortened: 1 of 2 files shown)


public/v4/apps/strapi.yml
```diff
@@ -33,7 +33,7 @@ caproverOneClickApp:
     variables:
         - id: $$cap_strapi_version
           label: Strapi Version
-          defaultValue: 3.5.3
+          defaultValue: 3.5.4
```
LiquidITGuy, line 36: will disappear after merging #395 (already merged on my master by mistake)

githubsaturn:
Thanks!

PR #359, opened 2021-02-07
[WIP] Add Penpot

sardaukar:
First of all, thank you for your contribution! 😄

### ☑️ Self Check before Merge

- [x] I have tested the template using the method described in README.md thoroughly
- [x] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [x] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [x] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [x] Icon is added as a png file to the logos directory.

Just need a little help with this template for Penpot (https://github.com/penpot/penpot)

Not sure what I'm doing wrong, but I get 

image

when deploying. The original docker-compose is [here](https://raw.githubusercontent.com/penpot/penpot/develop/docker/images/docker-compose.yaml)

Please help. Thanks!

PR #398, opened 2021-04-21
Add Penpot app. Closes #359

kgnfth:


timconsidine:
Thanks so much for packaging this app.
It's awesome.

I am however having a problem with exporting to PDF/PNG/JPG/SVG
I have restarted the penpot-exporter container but it fails consistently : UNEXPECTED ERROR OCCURRED
I have exported the file and loaded it into the hosted Penpot system, and it exports fine there.

Any thoughts on how to diagnose / resolve in the self-hosted version?

kgnfth:
@timconsidine Hello sir, unfortunately I can't help you as I do now own any servers anymore because servers are expensive
and I don't want to install any virtual machine junks on my machine to work offline.
I am really sorry.

PR #403, opened 2021-04-22
upgrade n8n from 0.67.3 to 0.116.1

LiquidITGuy:

### DIFF
--- a/public/v4/apps/n8n-io.yml
+++ b/public/v4/apps/n8n-io.yml
@@ -19,6 +19,8 @@ services:
         image: n8nio/n8n:$$cap_n8n_version
         volumes:
             - $$cap_appname-data:/root/.n8n
+            - $$cap_appname-data-write:/files
+            - $$cap_appname-docker-daemon:/var/run/docker.sock
         environment:
             DB_POSTGRESDB_DATABASE: n8n
             DB_POSTGRESDB_HOST: srv-captain--$$cap_appname-db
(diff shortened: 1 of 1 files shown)


public/v4/apps/n8n-io.yml
```diff
@@ -81,9 +84,7 @@ caproverOneClickApp:
         end: >-
             n8n is deployed and available as $$cap_appname.
 
-            IMPORTANT!! You need to follow these post-installation steps
-
-            First, please change your nginx config according to this issue https://github.com/caprover/one-click-apps/issues/265
```
LiquidITGuy, line 86: not needed anymore

githubsaturn:
Thanks!

public/v4/apps/n8n-io.yml
```diff
@@ -19,6 +19,8 @@ services:
         image: n8nio/n8n:$$cap_n8n_version
         volumes:
             - $$cap_appname-data:/root/.n8n
+            - $$cap_appname-data-write:/files
+            - $$cap_appname-docker-daemon:/var/run/docker.sock
```
githubsaturn, line 23: Yikes. Just saw this, it should be:
```
            - /var/run/docker.sock:/var/run/docker.sock
```

githubsaturn: I'll make the change

LiquidITGuy: thanks :)

PR #405, opened 2021-04-23
Upgrade nextcloud from 19.0.0 to 21.0.1

LiquidITGuy:
Upgrading nextcloud from 19 to last version (21.0.1)
Upgrading mariadb from 10.5.3 to last version (10.5.9)
Adding a redis cache to improve performances

I'm using this version on my server.
The server works, nextcloud report says it's fine, all my extensions too.
fits with all my usages but I do not tried all extensions (to many extensions on the market)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 dist/*
 node_modules
 .DS_Store
+.idea/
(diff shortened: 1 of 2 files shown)


githubsaturn:
Thank you!

PR #408, opened 2021-04-24
Upgrade Matomo from 3.11.0 to 4.2.1

LiquidITGuy:
Upgrade database dependencies
remove useless params and useless root password
upgrade Matomo from 3.11.0 to 4.2.1

### DIFF
--- a/public/v4/apps/matomo.yml
+++ b/public/v4/apps/matomo.yml
@@ -44,10 +44,12 @@ caproverOneClickApp:
           label: MariaDB Matomo User Password
           description: The password to use for the matomo database
           validRegex: /.{1,}/
+          defaultValue: $$cap_gen_random_hex(32)
         - id: $$cap_db_pass
           label: MariaDB Root Password
           description: The root password to use for the MariaDB instance
           validRegex: /.{1,}/
+          defaultValue: $$cap_gen_random_hex(32)
     instructions:
         start: Matomo is the leading open-source analytics platform that gives you more than powerful analytics.
         end: Matomo is successfully deployed!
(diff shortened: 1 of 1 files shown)


public/v4/apps/matomo.yml
```diff
@@ -8,11 +8,11 @@ services:
             MYSQL_DATABASE: matomo
             MYSQL_USER: matomo
             MYSQL_PASSWORD: $$cap_db_pass
-            MYSQL_ROOT_PASSWORD: $$cap_db_root_pass
```
githubsaturn, line 11: Why is this removed? Isn't it needed for mysqldump?
To make it easier for the user, we can use `$$cap_gen_random_hex(18)` as the default value.

githubsaturn (commented):

LiquidITGuy:
Thanks for the review

I removed those lines cause they are not on the matomo docker compose sample in the last version but it can be a misunderstanding by me ?

For the size of the random password I'll fix it :)

public/v4/apps/matomo.yml
LiquidITGuy: I think you can just use the matomo users (who already have all the right on the table on the database)
I removed the usage of the root account to prevent missuse of root account but I can put it back if you prefear ?

LiquidITGuy: Ps : I'm not sure about the fact the user can use mysql_dump, I'll check the matomo documentation for dump

githubsaturn: I see. I didn't know there is an existing user that has access to all the tables. That should be fine then.

PR #416, opened 2021-05-04
Corrected for spelling and RegEx validation

nativeit:
This template contained several spelling errors ("STMP" rather than "SMTP") and something with how the RegEx validation for the boolean SMTP entries was formatted caused the validation to fail using the default values. Adding " " around the default values in the definition file fixed this issue.

- [X ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [X ] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [X ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [X ] Icon is added as a png file to the logos directory.


githubsaturn:
This file is automatically generated from the yaml file you updated earlier. They shouldn't be manually touched.

nativeit:
Thanks, I'll remember that for the future. I appreciate everything you've done!

PR #422, opened 2021-05-05
Add Seafile (official image) support

D3rise:
This adds Seafile support with official image: `seafileltd/seafile-mc`


githubsaturn:
Thanks!

kgnfth:
correct me if i am wrong, but i don't see difference with #390 ?

D3rise:
Looks like #390 used official image, but for some reason didn't set `isOfficial` flag to `true`, so in some way (except that this PR don't use `memcached`) they are identical. I think, we need to rename it to "Seafile (without memcached)"

kgnfth:
> Looks like #390 used official image, but for some reason didn't set `isOfficial` flag to `true`, so in some way (except that this PR don't use `memcached`) they are identical. I think, we need to rename it to "Seafile (without memcached)"

Yes i agree

PR #424, opened 2021-05-05
Adds rstudio server 

tschannik:


tschannik:
I added explanation steps on how to configure the one-click-app to keep persistent data but was not able to get it to work. Any ideas in how to do this successfully?

public/v4/apps/rstudio.yml
```diff
@@ -0,0 +1,39 @@
...
+
+            Enter your RStudio Configuration parameters and click on next. It will take about a minute for the process to finish.
+        end: >-
+            Aaaand you're done! 🔥 
+            Your RStudio Server instance is available at http://$$cap_appname.$$cap_root_domain
+            Use username "rstudio" and your super secret passwort to login to your RStudio Server instance
+
+
+            To keep files after restarts you have to edit your caprover config at /captain/data/config-captain.json
+            and set "hasPersistentData": true
```
githubsaturn, line 35: You don't need this. All you need to do is to add a persistent directory whose files will be preserved during restarts, something like this:
```
        image: rocker/rstudio:$$cap_rstudio_version
        volumes:
            - $$cap_appname-code:/home/rstudio
```

Also, you may need to add `ROOT=TRUE` as an env var to be able to create and store files in this volume. Maybe, maybe not. Give it a shot either way.

public/v4/apps/rstudio.yml
```diff
@@ -0,0 +1,39 @@
...
+    displayName: RStudio
+    isOfficial: true
```
githubsaturn, line 37: I don't think Rocker is official. It's community maintained.

githubsaturn (commented):
Overall looks good, just some minor comments.

public/v4/apps/rstudio.yml
tschannik: Yep, you're right. Updated the value

public/v4/apps/rstudio.yml
tschannik: Works like a charm 👍 thanks

githubsaturn:
Thanks! Looks good! Just a quick validation error to fix:
```
Error: Description too long for rstudio  - keep it below 200 chars
```

tschannik:
@githubsaturn unlucky .. was 204 characters 😅 should be fine now

githubsaturn:
Thanks!

PR #428, opened 2021-05-07
Added Jellyfin

### LINKED ISSUE #201: App Suggestion: Jellyfin
Open source alternative to Plex.

I've already installed this manually, but I'm raising an issue to remind myself to create this as a one-click app.

Official Docker image here: https://hub.docker.com/r/jellyfin/jellyfin/

dramikei:
I have added one-click support for Jellyfin, a media streaming server like Plesk. I tested the app by uploading a video directly into the container's volume and playing it on my home pc.

This also resolves #201 Which has been open for almost one year now.

Jellyfin uses a different port for HTTP and HTTPS, caprover does provide an 'extra' parameter to map HTTP to the container's port but the same isn't available for HTTPS.

Enabling HTTPS in app config (from caprover dashboard) does allow the user to have HTTPS connection to Jellyfin, meaning the "Enable HTTPS" setting inside of jellyfin is useless and probably doesn't even work.

Note: Jellyfin support for Raspberry Pi may be limited because Caprover does not recognize 'device' parameters of the yml (As stated in the README), Though it should install just fine, I just don't have a Rpi lying around to test and be sure. If someone from the community tests and let me know that I will edit the description in the future.

- [x] I have ensured that I put as many default values as possible (except passwords) to ensure minimum effort required for end-users to get started.
- [x] I have made sure that instructions. start and instructions. end are clear and self-explanatory.


githubsaturn:
Thanks!

PR #432, opened 2021-05-10
added proxy pool

Ak4zh:


githubsaturn:
What's the use-case for this?

Ak4zh:
From the app description:
`The main function of the crawler agent IP pool project is to regularly collect the free agent verification and storage published on the Internet, and to verify the availability of the agent by regularly verifying the storage agent. It provides two ways of using API and CLI. `

It collects free open proxies and regularly verify to keep only active ones in the pool, which you can use in your scrapy or any other scrapping related projects.

PR #435, opened 2021-05-28
Add Invidious

Zottelchen:
This adds https://github.com/iv-org/invidious as one click app.

- [ ] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version. ❗ This is not possible, since only a 'latest' tag is provided (https://hub.docker.com/r/omarroth/invidious/tags) for Invidious.


public/v4/apps/invidious.yml
```diff
@@ -0,0 +1,58 @@
...
+                https_only: false
+                domain:
+        caproverExtra:
+            containerHttpPort: '3000'
+
+caproverOneClickApp:
+    variables:
+        - id: '$$cap_version'
+          label: Invidious Version
+          defaultValue: 'latest' # there is no other tag provided
```
githubsaturn, line 41: Can you try using the sha tag?

githubsaturn (commented):
Thank you for the PR!

public/v4/apps/invidious.yml
Zottelchen: I can do that, it works. (I am not entirely sure if it is better that way, since I imagine that invidious updates fairly often, because of changes youtube does.)

githubsaturn: Your comment in the variable label is pretty self explanatory. Thanks!

PR #437, opened 2021-06-02
Add NocoDB

christophemarois:

Tried it and it works beautifully.

image


githubsaturn:
Great, thanks!

ibarreda:
Without persistent data Noco DB is of little use. I tried to change the persistent directory path, but it doesn't work either. 

1) You install Noco DB , 2) You create a project 3) You go back to Caprover, save the app and you have a nice uninstalled app. 

It would be interesting to check this, that it is really stored persistently, it is a great app, but I find it not very useful if every time I have to configure the admin data and add projects.

Translated with www.DeepL.com/Translator (free version)

githubsaturn:
According to this docker-compose example, NocoDB is a stateless app: https://github.com/nocodb/nocodb/blob/master/docker-compose/mysql/docker-compose.yml#L19-L31

It doesn't have any persistent volume. All the data is stored on the db.

PR #454, opened 2021-07-10
Add Transmission

amjed-ali-k:
qBittorent in One-click-Apps is not working for me. So I created Transmission 


public/v4/apps/transmission.yml
```diff
@@ -0,0 +1,63 @@
...
+            PGID: '1000'
+            TZ: $$cap_tz
+            TRANSMISSION_WEB_HOME: /$$cap_ui/
+            USER: $$cap_user
+            PASS: $$cap_pass
+            WHITELIST: '' #optional
+            HOST_WHITELIST: '' #optional
+            WEBUI_PORT: '9091'
+        ports:
+            - 9091:9091
```
githubsaturn, line 16: You probably don't need to port map 9091 as it's already being using proxied through nginx.

public/v4/apps/transmission.yml
```diff
@@ -0,0 +1,63 @@
...
+          description: Path to pictures folder where u want to save the downloaded torrents. You can mention an existing directory here too. Eg. /home/user/torrents/
+          validRegex: '/.{1,}/'
+        - id: $$cap_ui
+          label: User Interface
+          defaultValue: transmission-web-control
+          description: Specify an alternative UI options are combustion-release, transmission-web-control, kettu and flood-for-transmission.
+          validRegex: /^(combustion-release|transmission-web-control|kettu|flood-for-transmission)$/
+    instructions:
+        start: >-
+            Transmission is designed for easy, powerful use. Transmission has the features you want from a BitTorrent client: encryption, a web interface, peer exchange, magnet links, DHT, µTP, UPnP and NAT-PMP port forwarding, webseed support, watch directories, tracker editing, global and per-torrent speed limits, and more.
```
githubsaturn, line 55: Probably need to mention what ports are being mapped to make sure the user knows about them.
- They need to be not used.
- They need to be open and accessible from their firewall

public/v4/apps/transmission.yml
amjed-ali-k: sure

public/v4/apps/transmission.yml
amjed-ali-k: sure

githubsaturn:
Thank you!

PR #471, opened 2021-07-28
Added Prowlarr

kgnfth:


public/v4/apps/prowlarr.yml
```diff
@@ -0,0 +1,53 @@
...
+            PGID: $$cap_guid
+        volumes:
+            - $$cap_appname-config:/config
+        caproverExtra:
+            containerHttpPort: '9696'
+caproverOneClickApp:
+    variables:
+        - id: $$cap_pr_version
+          label: Prowlarr Version
+          defaultValue: 'sha256:958269a836506d87d2fc807d7e86dfada1191d994757546c35dc9a785a3f6a37'
```
githubsaturn, line 19: Do they not support human readable tags? Looks like they do have things like this: `arm64v8-nightly-0.1.0.688-ls79`

SHA digests are the last resource in case a human readable tag is not available.

kgnfth: Yes but i thought i would use sha256 because they release it nightly. Anyway i will revert to tagged version

PR #478, opened 2021-08-14
Sea file pro template

anaganisk:
created seafile pro yaml file

- [ ☑️] I have tested the template using the method described in README.md thoroughly
- [☑️ ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [☑️] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version. Seafile pro only realases latest tag
- [☑️ ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [☑️ ] Icon is added as a png file to the logos directory.


kotx:
Has anybody managed to deploy and/or upgrade to Seafile Pro on Caprover?

anaganisk:
I did with the original PR, I couldn't find time with the issue presented by dumrop. Its still running for me

PR #481, opened 2021-08-23
Fix docker hub ref for firefly

Church13:

The official docker hub image for Firefly III has moved -- this PR fixes the ref.


Church13:
As per https://hub.docker.com/r/jc5x/firefly-iii -- the old jc5x/firefly-iii hub image has been deprecated, and Firefly III is moving to fireflyiii/core (https://hub.docker.com/r/fireflyiii/core).

githubsaturn:
Have you tested this PR? Looks like the default tag is no longer available under `fireflyiii/core`.

Church13:
> Have you tested this PR? Looks like the default tag is no longer available under `fireflyiii/core`.

I did test it, but manually specified the tag on my install -- corrected to reflect a recent, available tag in the fireflyiii/core docker hub. Good catch.

githubsaturn:
Perfect, thanks!

PR #488, opened 2021-09-05
Fixed db name

Fabian9799:
Fixed a small typo where the database name was set to `yagpdb_prod` instead of `yagpdb_production` like in postgres


githubsaturn:
Thank you!

PR #492, opened 2021-09-13
New App: iRedMail - Open Source Mail Server

ido-pluto:
#### This is a custom build of iRedMail.

I added some fixes to make this work with CapRover + Some adjustment I think is needed.

Changes:
- Remove the example certificate from the web app so it works with CapRover HTTPS
- Added link to the Webmail from the iRedMail Admin
- Added DNS records in the 'Edit domain' page
- Option to change the installed version of iRedMail Admin (On the first build, not work after docker image built)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "fs-extra": "^9.0.1",
     "gh-pages": "^3.1.0",
-    "prettier": "^2.0.5",
+    "prettier": "^2.4.1",
     "yaml": "^1.10.0"
   }
 }


githubsaturn:
This is very customized and very complex. I don't think it's maintainable in its current form. I suggest either make it a tad more maintainable my moving the changes upstream (to the source code) or remove the customizations.

idoio:
True, but that way, my docker file will work with every update of the original docker image without the need to do anything :)

btw: if the build doesn't work after formatting with prettier, I wilt upload it on JSON format

githubsaturn:
> but that way, my docker file will work with every update of the original docker image without the need to do anything :)

Not sure if I understand. Are you saying the current method works with every update moving forward?

Let me rephrase my question. Why wouldn't you want to move these changes to the main repo? Why do we need to have these customization in CapRover one click app, but not the default (non-CapRover) installation?

idoio:
Yes - spouse too. I don't move the changes because there are some changes that make iRedMail compatible with CapRover.
If you install without CapRover, you don't need those.

I guess that all of my changes are necessary because in 'one-click-app' you are not expected to go to the shell.

For example, the "DNS records" changes are compatible only with ubuntu, so there is no point to push them to the main repo.
I am not trying to change iRedMail, just trying to make it compatible and comfortable to use with CapRover.

githubsaturn:
I see. Okay, let's do this.


Regarding formatting, feel free to revert to yaml. I'll merge and then will sort out the prettier issue.

PR #517, opened 2021-10-15
add kimai2 to one click apps

lcx:

### DIFF
--- a/kimai2.yml
+++ b/kimai2.yml
@@ -31,7 +31,7 @@ services:
         caproverExtra:
             notExposeAsWebApp: 'true'
     $$cap_appname:
-        image: lcxat/nginx-fpm-reverse-proxy
+        image: lcxat/nginx-fpm-reverse-proxy:$$cap_nginx_version
         environment:
             FPM_NGINX_HOST: srv-captain--$$cap_appname-fpm
         volumes:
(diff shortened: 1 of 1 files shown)


kimai2.yml
```diff
@@ -0,0 +1,95 @@
...
+          defaultValue: $$cap_gen_random_hex(32)
+    instructions:
+        start: |-
+           With Kimai, the boring process of feeding Excel spreadsheets with your working hours is not only simplified, it also offers dozens of other exciting features that you don't even know you're missing so far!
+        end: |-
+            Aaaand you're done! 😄
+            Your service is available at http://$$cap_appname.$$cap_root_domain
+            -------------------------------------------------------------------
+    displayName: Kimai
+    isOfficial: true
```
githubsaturn, line 93: This is not the official image. It's from a 3rd party repo: https://github.com/tobybatch/kimai2
The original repo is here: https://github.com/tobybatch/kimai2

kimai2.yml
```diff
@@ -0,0 +1,95 @@
...
+            MYSQL_USER: $$cap_mariadb-user
+            MYSQL_PASSWORD: $$cap_mariadb-pass
+            MYSQL_ROOT_PASSWORD: $$cap_maridb-rootpass
+        volumes:
+            - $$cap_appname-db-data:/var/lib/mysql
+        restart: unless-stopped
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+    $$cap_appname:
+        image: lcxat/nginx-fpm-reverse-proxy
```
githubsaturn, line 34: This is using an implicit `latest` tag. Let's fix the version to prevent it from breaking the installations with future updates.

githubsaturn (commented):
Thanks for the PR! Just a few minor comments.

kimai2.yml
lcx: oh yes, you are right. Kimai does refer to the third party link https://www.kimai.org/documentation/docker.html but never states it's the "official".

kimai2.yml
lcx: yep, never considered to check the :latest on my own containers. Let me fix that as well.

lcx:
@githubsaturn both issues fixed. ✅

PR #550, opened 2021-10-31
Update default version and add BASE_URL

jfowl:
I updated the tag to a newer version and inserted a sane BASE_URL env, to reduce needed configuration.


githubsaturn:
Thanks for the PR. Please run `npm run formatter-write` to fix the formatting issue. The indentation is incorrect. Alternatively, you can copy-paste from another app that has env var section

githubsaturn:
Thanks!

PR #560, opened 2021-11-10
Added CloudFlare DDNS

Inoriol:
There is no "fixed" version of this container, so you have to use "latest" tag

- [❌] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.


public/v4/apps/cloudflareddns.yml
```diff
@@ -0,0 +1,93 @@
...
+            RRTYPE: $$cap_rrtype
+            DNS_SERVER: $$cap_dns_server
+            CRON: $$cap_cron
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+caproverOneClickApp:
+    variables:
+        - id: $$cap_api_key
+          label: Cloudflare API token
+          defaultValue: 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklm'
```
githubsaturn, line 19: Leave the default value empty as this is just an example.

githubsaturn (commented):

githubsaturn:
For future ref: looks like this image does not have any proper versioning, hence we have to use the `latest` tag.

githubsaturn:
Thanks @Inoriol !

PR #571, opened 2021-11-20
Fixed package.json formatter script

Skayo:
The script didn't work with single quotes.
When I ran the script, I got:
```
'yml)'' is not recognized as an internal or external command
```
Using double quotes works, tho.
Maybe it's just an issue on Windows, but I checked and most people use double quotes for this exact prettier command.

Also I have added a link to the README for a custom one-click-app repository example. I can remove it again if you don't want it.


githubsaturn:
Thanks!

PR #575, opened 2021-11-29
Invoice Ninja

VeselaHouba:


public/v4/apps/invoiceninja.yml
```diff
@@ -0,0 +1,150 @@
...
+            notExposeAsWebApp: true
+
+    $$cap_appname-server:
+        restart: always
+        depends_on:
+            - $$cap_appname-app
+        environment:
+            APP_CONTAINER: srv-captain--$$cap_appname-app:9000
+        volumes:
+            - $$cap_appname-app-public:/var/www/app/public:ro
```
githubsaturn, line 127: Remove `:ro`, it's not consumed by CapRover.

public/v4/apps/invoiceninja.yml
```diff
@@ -0,0 +1,150 @@
...
+        caproverExtra:
+            # Workaround for hard-coded container name in nginx config
+            dockerfileLines:
+                - FROM nginx
```
githubsaturn, line 131: Add a version tag here. No need to export it in variables, but at least we should fix this version tag.

public/v4/apps/invoiceninja.yml
```diff
@@ -0,0 +1,150 @@
...
+                - ADD https://raw.githubusercontent.com/invoiceninja/dockerfiles/master/config/nginx/in-vhost.conf /etc/nginx/templates/in-vhost.conf.template
```
githubsaturn, line 132: This is always pointing to the latest `master` file. Instead, use a fixed commit for this link.

public/v4/apps/invoiceninja.yml
githubsaturn: Same with `:rw` in other lines.

githubsaturn (commented):
Looks good, I'm excited for this!

Just to confirm, this is tested and it's working?

public/v4/apps/invoiceninja.yml
VeselaHouba: Makes sense. If they do some breaking changes in next release, the template will need to be reviewed anyway.

VeselaHouba:
> 

Yes, it seems to be working. I have only tested to the extent of creating example company and checking that my changes are persistent.

public/v4/apps/invoiceninja.yml
VeselaHouba: fixed

public/v4/apps/invoiceninja.yml
VeselaHouba: fixed to 1.21

public/v4/apps/invoiceninja.yml
VeselaHouba: fixed to latest tag 5.3.34

VeselaHouba:
Rebased & squashed. @githubsaturn It's all yours now :)

githubsaturn:
Thanks! Looks great!

PR #580, opened 2021-12-17
Remove maxmind container from Plausible

sergioisidoro:
## This is a heavily opinionated PR. Feel free to close. Rational in the description

- Since early 2020 Maxmind changed their licensing. This means that if you want to deploy Plausible for anything commercial, or in support of anything commercial, you need to pay.

- Since then Plausible started shipping with scripts to fetch data from a free source (db-ip): https://github.com/plausible/analytics/blob/1dba113e2fe5fb3593c56fa6994f6d088d3fec80/config/runtime.exs#L110

- This app has mandatory fields for Maxmind api call, although it is not strictly necessary anymore

- This makes me think it's better to default to a free installation of plausible, and allow users to add the maxmind container if they wish to do so with a custom template.

- I tried check if there's any way to add / remove containers with conditional logic, but I couldn't find anything for that. This would be a great use case for basic `if/else` templating in the apps :)

### DIFF
--- a/public/v4/apps/plausible.yml
+++ b/public/v4/apps/plausible.yml
@@ -37,7 +37,7 @@ services:
         caproverExtra:
             containerHttpPort: '8000'
             dockerfileLines:
-                - FROM plausible/analytics:v1.1.1
+                - FROM plausible/analytics:$$cap_PLAUSIBLE_VERSION
                 - CMD ["sh", "-c", "sleep 10 && /entrypoint.sh db createdb && /entrypoint.sh db migrate && /entrypoint.sh db init-admin && /entrypoint.sh run"]
         environment:
             ADMIN_USER_EMAIL: $$cap_ADMIN_USER_EMAIL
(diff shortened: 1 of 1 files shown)


sergioisidoro:
Ping @thomascenni as they were the original authors of this app. What do you think?

thomascenni:
OK for me, thanks for pinging!
I am not the owner of the repository :)
I think you could also update the version of the other components like yandex/clickhouse-server and postgresql if you want.
Thomas

sergioisidoro:
I know @thomascenni , but since you were the author, I thought you would like a say in the change :) I will also add postgres and clickhouse versions as a variable to make it easier to upgrade when new versions come up.

githubsaturn:
Thanks for the PR! I have no idea what Plausible is and I have never worked with it. So I trust your judgement and @thomascenni  on this.

Just one quick suggestion: makes sense to add these information to instructions.start so that users will know about these specific details.

sergioisidoro:
@githubsaturn Done. I hope I understood your request :)

githubsaturn:
Perfect, thank you!

PR #581, opened 2021-12-20
Added app Redmine (with both PostgreSQL & MySQL) and corrected a minor error with URL for upshift/dolibarr tags

nativeit:

I tested Redmine with both database options to ensure that it deploys fully, without errors, and remains stable over the course of 24-hours. All default values are based off of the recommendations of the official documentation on the Docker repositories. Redmine and both databases are all pulling from official images. It was somewhat difficult to find a reasonably clear and sharp Redmine logo, so I made a 1:1 recreation of it in Illustrator that matches their official logo in shape, position, logotype, and color.


githubsaturn:
Great, thank you!

public/v4/apps/redmine-mysql.yml
```diff
@@ -0,0 +1,70 @@
+captainVersion: 4
+services:
+    $$cap_appname-db:
+        image: mysql:$$cap_mysql_version
+        command: --default-authentication-plugin=mysql_native_password
```
githubsaturn, line 5: This is being ignored by CapRover. But if you tested it, I assume this line isn't required.

nativeit: Yeah, I simply copied that from MySQL's official Docker docs, there wasn't any specific intent involved with it.

PR #589, opened 2021-12-24
Introducing Chadburn scheduler

maietta:

- [✓] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.

Hello Caprover team,

I maintain an open source project called Chadburn and have tested it extensively in Caprover over the last few months. I am finally comfortable with the inclusion of Chadburn as an official app in the Caprover Apps Repository.

I am glad to be able to finally offer my first contribution to the Caprover community of users.

Thank you.


public/v4/apps/chadburn.yml
```diff
@@ -0,0 +1,33 @@
...
+version: 3.3
+services:
+    '$$cap_appname':
+        volumes:
+            - /var/run/docker.sock:/var/run/docker.sock:ro
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+            dockerfileLines:
+                - FROM premoweb/chadburn:$$cap_version
+                - CMD ["daemon"]
```
githubsaturn, line 11: Looks like the default CMD is `daemon`. 
https://github.com/PremoWeb/chadburn/blob/9db334edebdb034c26c5f0c0ead57226930c37a3/Dockerfile#L18

If that's correct, then we don't need to override it here. Instead of `dockerfileLines`, you can simple add an `image` field (sibling to `volume` a few lines before this). See here for an example:
https://github.com/caprover/one-click-apps/blob/0ad28e01b57eccb78d5b53ea2901844ec4736c6c/public/v4/apps/mysql.yml#L3-L6

githubsaturn (commented):
Thanks! Looks good! Just one minor comment.

public/v4/apps/chadburn.yml
maietta: Updated! Thank you.

maietta:
Sharp eye. Thanks for the suggested change(s). I have re-tested my changes and committed those changes.

Listed here, I now see the reference 5233ddb which is my most recent commit from a few minutes ago. I think we're good to go.

maietta:
I'm closing this due to confusion of my steps taken. I will be preforming a new pull request shortly.

PR #592, opened 2021-12-27
[NEW] Mercure one-click template

fgarioli:

- [v] I have tested the template using the method described in README.md thoroughly
- [v] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [v] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [v] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [v] Icon is added as a png file to the logos directory.


githubsaturn:
Thank you!

PR #593, opened 2022-01-01
Update chadburn.yml (Now supports multiple CPU architectures)

maietta:
Chadburn now support multiple CPU architectures with version 1.0.1. Also removed the "v" prefix for all releases.


githubsaturn:
Thanks!

PR #594, opened 2022-01-03
Appsmith with database

aqidd:


githubsaturn:
Thanks for the PR! 

Just one question: the Docker Compose file in their [official guide](https://docs.appsmith.com/setup/docker) has a volume at `/appsmith-stacks` - was this removed on purpose?

aqidd:
Ah you're right. Upon further checks, this volume is useful when you do a db dump. Will update the definition with volume

aaron5670:
Looks great! 👍

githubsaturn:
Thank you!

PR #598, opened 2022-01-08
bump kc version to current, autogen passwords, fixes #597

### LINKED ISSUE #597: App issue: Keycloak
Sorry for sending a bug without a fix, but I'm attempting to deploy Keycloak from the one-click-apps config.

There are two issues:

1) The keycloak default version in OCA is `6.0.1`.  The current Keycloak release is `16.0.0`.  This being security software, I'm willing to bet there are some important fixes in there.

2) When attempting to deploy the OCA Keycloak with the latest tagged docker version (by overriding the field in the OCA launch wizard) `16.1.0`, it doesn't start.

The fix is:

1) Figure out what needs to be changed in the OCA config to make 16.x work

2) Bump the version

I will attempt to work on this if I can find the time, but I wanted to create an issue to track it as it's probably insecure/vulnerable in the default config (deploying `6.0.1`).

sneak:

### DIFF
--- a/public/v4/apps/keycloak.yml
+++ b/public/v4/apps/keycloak.yml
@@ -1,3 +1,4 @@
+# tested on CR 1.10.1 2022-01-08 by sneak
 captainVersion: 4
 services:
     $$cap_appname-db:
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thank you!

PR #601, opened 2022-01-14
Fix Minio version variable

dangxunb:

### DIFF
--- a/public/v4/apps/minio.yml
+++ b/public/v4/apps/minio.yml
@@ -14,7 +14,7 @@ services:
         caproverExtra:
             containerHttpPort: '9001'
             dockerfileLines:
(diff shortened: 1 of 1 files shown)


public/v4/apps/minio.yml
```diff
@@ -14,7 +14,7 @@ services:
         caproverExtra:
             containerHttpPort: '9001'
             dockerfileLines:
-                - FROM minio/minio:RELEASE.2021-08-17T20-53-08Z
+                - FROM minio/minio:$$cap_minio_version
```
githubsaturn, line 17: Thanks for the PR!

I'll change the default version to `RELEASE.2021-08-17T20-53-08Z` after merging since that is the tested version.

Issue #583, opened 2021-12-21
App Suggestion: Corteza CRM Salesforce alternative


jumus-jumbuck:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
Corteza CRM Salesforce alternative

**Do you know if there is any official Docker image for the app?**
https://github.com/cortezaproject/corteza-server
cortezaproject.org

PR #603, opened 2022-01-18
Add Corteza #583

amjed-ali-k:

#583


aaron5670:
Looks good! 💯

public/v4/apps/corteza.yml
```diff
@@ -0,0 +1,63 @@
+captainVersion: 4
+
+services:
+    $$cap_appname:
+        image: cortezaproject/corteza:$$corteza_version
+        restart: always
+        depends_on:
+            - $$cap_appname-db
+        volumes:
+            - $$cap_appname-data:/data"
```
githubsaturn, line 10: Extra `"` at the end

public/v4/apps/corteza.yml
```diff
@@ -0,0 +1,63 @@
...
+        environment:
+            VIRTUAL_HOST: $$cap_appname.$$cap_root_domain
+            DB_DSN: postgres://$$cap_postgre_username:$$cap_postgre_password@srv-captain--$$cap_appname-db:5432/$$cap_postgre_db_name?sslmode=disable
+            HTTP_WEBAPP_ENABLED: true
+            DOMAIN: $$cap_appname.$$cap_root_domain
+            VERSION: $$corteza_version
+    $$cap_appname-db:
+        image: postgres:13
```
githubsaturn, line 18: This app needs persistent volume, otherwise after a restart all data will be lost.

public/v4/apps/corteza.yml
```diff
@@ -0,0 +1,63 @@
...
+            Corteza has been successfully deployed!
+            --------------------------------------------
+            Before you proceed, please enable Websocket Support
+            https://captain.$$cap_root_domain/#/apps/details/$$cap_appname
+            --------------------------------------------
+            App is available as http://$$cap_appname.$$cap_root_domain
+    displayName: Corteza
+    isOfficial: true ## Only if all images used here are official or from a trusted source.
+    description: Corteza is a free, open-source, Low Code platform for building your organisation key applications, growing its productivity and protecting its data and the privacy of all those concerned.
+    documentation: This docker-compose is taken https://github.com/knadh/listmonk
```
githubsaturn, line 63: Wrong link? `listmonk` is a totally different app, right?

githubsaturn (commented):
Thanks! Just a few minor issues.

amjed-ali-k:
Fixed 😺

public/v4/apps/corteza.yml
amjed-ali-k: done

githubsaturn:
Thanks!

nativeit:
I have encountered an issue, it's likely something fairly simple to address. After initial deployment, I enabled https, enabled web sockets as directed, and was able to create my first user and login. But before I could then begin setting up the various components of Corteza, I encountered a blank white screen with a mixed content error indicated in the console:
```
Mixed Content: The page at 'https://corteza.domain.tld/auth/callback?code=OWVMYJU3MZMTM2NMYY0ZODLMLTHLNMUTZMJMNTGYNJYZNJNI&state=m3fp8kkfvy' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://corteza.domain.tld/auth/oauth2/default-client'. This request has been blocked; the content must be served over HTTPS. 
```

This is usually due to a simple missing `s` within a config file, and I am currently looking into it. I'll update this if/when I find the root cause, and if I can identify how to fix it I can make the change under a new PR or post back here for you to handle, whichever you'd prefer.

nativeit:
I have found one environment variable that may be relevant (`HTTP_CLIENT_TSL_INSECURE`), I am following the docs at https://docs.cortezaproject.org/corteza-docs/2020.12/dev-ops-guide/server-configuration/index.html

It indicates the use of a `.env` file paired with the `docker-compose.yml`, so if they haven't build their Docker image to support all of the flags in this doc, (@githubsaturn correct me if I'm wrong here) it may need to be implemented as a Dockerfile line. 

Just for example:
```
        caproverExtra:
            dockerfileLines:
                - FROM cortezaproject/corteza:$$corteza_version
                - ENV HTTP_CLIENT_TSL_INSECURE=$$cap_corteza_ssl
```

Also, while we're at it, the `postgresdb` should probably include a block disabling it as a webapp, ie:
```
        caproverExtra:
            notExposeAsWebApp: 'true'
```

githubsaturn:
Why not add it explicitly as an env var?
```diff
captainVersion: 4

services:
    $$cap_appname:
        image: cortezaproject/corteza:$$corteza_version
        restart: always
        depends_on:
            - $$cap_appname-db
        volumes:
            - $$cap_appname-data:/data
        environment:
            VIRTUAL_HOST: $$cap_appname.$$cap_root_domain
++          HTTP_CLIENT_TSL_INSECURE: $$cap_corteza_ssl
            DB_DSN: postgres://$$cap_postgre_username:$$cap_postgre_password@srv-captain--$$cap_appname-db:5432/$$cap_postgre_db_name?sslmode=disable
....
```

nativeit:
Wasn't initially sure that would provide the same result, but I'm testing it now. Corteza has a rather unique implementation and ambiguous documentation with some of these environment variables. I got it all working in one instance by removing `VIRTUAL_HOST` and explicitly setting `DOMAIN` as https://app.domain.tld, but in another case that just resulted in malformed URL's when calling the OAUTH endpoint. I'll figure it out and share what ultimately works.

I realize there are probably a not insignificant amount of people running Caprover on local development servers who might not need/want SSL in every case, so providing this as an option can sometimes be a little clunky depending on how the Docker image is configured. Let me just say that I _really_ appreciate developers who include a `ENABLE_SSL` flag, or something similar so we can just encode a boolean into our options. It seems Corteza has no such feature.

nativeit:
Oh, this is even more annoying than I imagined. If I include the `https://` in the `DOMAIN` prior to initial setup/login, then I end up with malformed URLS like `http://https//app.domain.tld/oauth/...`. So I remove the `https://` from `DOMAIN` and it forwards to SSL and allows initial registration no problem, then the original issue with blank screen and `MIXED CONTENT` errors occurs. Now if I add `https://` to `DOMAIN` specifically at this point, everything works a treat, all the OAUTH endpoints are happy and it's good to go. I'm going to continue looking for a workaround, otherwise this seems like an issue on Corteza's end of things. They don't seem to have properly considered how to implement this Docker image behind a reverse proxy with SSL.

nativeit:
Good news, this turned out to be an easy fix, despite ambiguity in Corteza's documentation. Everything can be left more or less the same, we just need to add the following environment variable to enable https support:
```
    LETSENCRYPT_HOST: $$cap_appname.$$cap_root_domain
```
I tested this and everything functioned more or less as expected. The app checks for this flag, and assumes SSL is enabled if it is set. I'm going to run it for another few days and make sure it can be deployed `null` without interfering with non-https operation. If @amjed-ali-k doesn't have an opportunity to update the repo by then, I'll open a PR with my own updates. Let me know if there's anything I should consider in the meantime. Thanks!

amjed-ali-k:
I guess `LETSENCRYPT_HOST` env only needs to be set if a user is planning to enable SSL and `null` if they are not. Then we might need to ask the user to input the host domain for `LETSENCRYPT_HOST` env.

@nativeit  Also kindly check the plugins when you are testing. It seems the plugin file needs to be persistent.
https://docs.cortezaproject.org/corteza-docs/2021.9/devops-guide/server-plugins/index.html

amjed-ali-k:
#615 Created a new pull request.

PR #627, opened 2022-02-22
Update paperless-ng.yml for proper gotenberg support

borjarobert:
Only version 6 of gotenberg works properly with paperless-ng. If the image name is left undefined (as it is right now) it downloads and deploys version 7.5, which gives a 404 error when called by paperless-ng. Setting it to gotenberg:6 fixes the issue.


githubsaturn:
Thanks!

PR #634, opened 2022-03-12
updated Calendso docker link

vigasdeep:

### DIFF
--- a/public/v4/apps/calendso.yml
+++ b/public/v4/apps/calendso.yml
@@ -2,7 +2,7 @@ captainVersion: 4
 services:
     # Calendso
     $$cap_appname:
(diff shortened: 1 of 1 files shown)


public/v4/apps/calendso.yml
```diff
@@ -2,7 +2,7 @@ captainVersion: 4
 services:
     # Calendso
     $$cap_appname:
-        image: ctadeu/calendso:$$cap_app_version
+        image: calendso/calendso:$$cap_app_version
```
githubsaturn, line 5: This image doesn't have any tags. Have you tested this change?

githubsaturn:
https://github.com/caprover/one-click-apps/issues/705

PR #638, opened 2022-03-27
Add Appwrite

danielrataj:
Please add Appwrite to one-click apps. Thank you very much ❤️ 
https://appwrite.io/


public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
+captainVersion: 4
+
+services:
+    $$cap_appname-redis:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'bitnami/redis:6.0.8-debian-10-r6'
+        ports:
+            - '6379:6379'
```
githubsaturn, line 9: Does the port need to be exposed to the public? If it's only being used by AppWrite, then, probably it won't need to be exposed.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            ALLOW_EMPTY_PASSWORD: 'yes'
+        volumes:
+            - $$cap_appname-redis-data:/bitnami/redis/data
+
+    $$cap_appname-mariadb:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'docker.io/bitnami/mariadb:10.7'
+        ports:
+            - '3306:3306'
```
githubsaturn, line 21: Same question about port.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+        environment:
+            MARIADB_ROOT_PASSWORD: $$cap_APP_DB_ROOT_PASS
+            MARIADB_DATABASE: $$cap_APP_DB_SCHEMA
+            MARIADB_USER: $$cap_APP_DB_USER
+            MARIADB_PASSWORD: $$cap_APP_DB_PASS
+        command: 'mysqld --innodb-flush-method=fsync'
```
githubsaturn, line 27: CapRover doesn't consume `command` as per docs here: https://github.com/caprover/one-click-apps#services

Either remove or convert to `dockerfileLines` if needed (search repo for examples)

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+        volumes:
+            - $$cap_appname-mariadb-data:/bitnami/mariadb
+
+    $$cap_appname-influxdb:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'appwrite/influxdb:1.0.0'
+        ports:
+            - '8086:8086'
```
githubsaturn, line 36: Same port question.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            _APP_MAINTENANCE_RETENTION_EXECUTION:
+            _APP_MAINTENANCE_RETENTION_ABUSE:
+            _APP_MAINTENANCE_RETENTION_AUDIT:
+            DOCKERHUB_PULL_USERNAME: $$cap_DOCKERHUB_PULL_USERNAME
+            DOCKERHUB_PULL_PASSWORD: $$cap_DOCKERHUB_PULL_PASSWORD
+            DOCKERHUB_PULL_EMAIL: $$cap_DOCKERHUB_PULL_EMAIL
+
+    $$cap_appname-appwrite-realtime:
+        image: 'appwrite/appwrite:$$cap_APP_VERSION'
+        entrypoint: realtime
```
githubsaturn, line 137: Similar to `command`, `entrypoint` is not consumed by CapRover. Please read docs and if needed convert to `dockerfileLines`

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            _APP_MAINTENANCE_RETENTION_AUDIT:
+            DOCKERHUB_PULL_USERNAME: $$cap_DOCKERHUB_PULL_USERNAME
+            DOCKERHUB_PULL_PASSWORD: $$cap_DOCKERHUB_PULL_PASSWORD
+            DOCKERHUB_PULL_EMAIL: $$cap_DOCKERHUB_PULL_EMAIL
+
+    $$cap_appname-appwrite-executor:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'appwrite/appwrite:$$cap_APP_VERSION'
+        entrypoint: executor
```
githubsaturn, line 218: Again, not consumed by CapRover.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            _APP_MAINTENANCE_RETENTION_AUDIT:
+            DOCKERHUB_PULL_USERNAME: $$cap_DOCKERHUB_PULL_USERNAME
+            DOCKERHUB_PULL_PASSWORD: $$cap_DOCKERHUB_PULL_PASSWORD
+            DOCKERHUB_PULL_EMAIL: $$cap_DOCKERHUB_PULL_EMAIL
+
+    $$cap_appname-appwrite-worker-database:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'appwrite/appwrite:$$cap_APP_VERSION'
+        entrypoint: worker-database
```
githubsaturn, line 304: Not consumed by CapRover.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            _APP_MAINTENANCE_RETENTION_AUDIT:
+            DOCKERHUB_PULL_USERNAME: $$cap_DOCKERHUB_PULL_USERNAME
+            DOCKERHUB_PULL_PASSWORD: $$cap_DOCKERHUB_PULL_PASSWORD
+            DOCKERHUB_PULL_EMAIL: $$cap_DOCKERHUB_PULL_EMAIL
+
+    $$cap_appname-appwrite-worker-builds:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'appwrite/appwrite:$$cap_APP_VERSION'
+        entrypoint: worker-builds
```
githubsaturn, line 385: Not consumed by CapRover.

public/v4/apps/appwrite.yml
```diff
@@ -0,0 +1,1338 @@
...
+            _APP_MAINTENANCE_RETENTION_AUDIT:
+            DOCKERHUB_PULL_USERNAME: $$cap_DOCKERHUB_PULL_USERNAME
+            DOCKERHUB_PULL_PASSWORD: $$cap_DOCKERHUB_PULL_PASSWORD
+            DOCKERHUB_PULL_EMAIL: $$cap_DOCKERHUB_PULL_EMAIL
+
+    $$cap_appname-appwrite-worker-audits:
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        image: 'appwrite/appwrite:$$cap_APP_VERSION'
+        entrypoint: worker-audits
```
githubsaturn, line 466: Not consumed by CapRover.

githubsaturn (commented):
Thanks for the PR. I left  a couple of issues.

Most importantly, please note that only a handful of variables are consumed by CapRover ([docs](https://github.com/caprover/one-click-apps#services)). If other variables, such as `entrypoint` is actually needed, you'll have to use `dockerfileLines`

public/v4/apps/appwrite.yml
danielrataj: I did remove the ports section.

public/v4/apps/appwrite.yml
danielrataj: I used dockerfileLines properly. I forgot about it. 😞

danielrataj:
It has all been fixed. Thanks for the hints. I really appreciated it. :-)

githubsaturn:
Thanks!

nativeit:
**Environment**:
- **Host**: OVHcloud Dedicated Bare Metal
- **CPU**: Intel Xeon-E 2136 12-Core
- **Mem**: 32GB
- **KVM**: Proxmox VE 7
- **VM**: 4-cores, 8GB mem
  - **OS**: Debian GNU/Linux 11 (bullseye)
  - **Kernel**: 5.10.0-13-amd64
  - **Docker** version 20.10.15, build fd82621
  - **Caprover** version 1.10.1

---
First let me say, as someone who has contributed a few apps to Caprover's collection, thanks and my compliments to @danielrataj for the immense effort this must have required relative to most other templates. It is understandably massive, with 17 individual service containers and a litany of environment variables, but somehow despite this I got it to deploy successfully on my first attempt, and for a while it ran just fine. Unfortunately, I think there's a flaw, and I think it may be almost unavoidable as it currently stands. I am filing this issue with the hopes that there may be a fairy simple solution.

The issue appeared the day after I deployed Appwrite and finished all of the initial setup, including the creation of a project and arranging for logs to be sent to Sentry. But at some point this evening, I noticed my app server was quickly running out memory and disk space, and when I checked I saw a cascade of Appwrite services failing, stopping, and re-spawning. When one service would fail and stop, it would cause an error in another which would fail and stop, and so on. Even when sending `STOP` commands to the containers, they would just spin back up and continue the death spiral.

I'd be curious to know if anyone else has experienced this? If so, I would like to evaluate the template to see if there is anything we can do to tweak the `depends_on` and `restart` entries for each of the individual services in an effort to encourage better reliability and a more controlled failure mode. I also thought there may be an approach that combines some of these supporting services to reduce the overall number of dependent containers, but I understand that, current issue aside, running services in containers 1:1 is fundamentally how Docker is intended to operate. But as it stands, with so many interdependent services, once the first domino falls it does not take long for this app to tank even a generously spec'd server altogether, along with any other unrelated apps, the Docker engine, and Caprover itself.

I am happy to participate and assist with the legwork. I have spun up a separate independent VM with its own Caprover environment so that I can work with Appwrite's template without affecting any other apps. I am currently working on writing a Bash script to speed up the deployment process, since entering so many environment variables one at a time using the usual one-click-app deployment interface, and then deleting each service individually, can be very time consuming, especially if one intends to run a series of test deployments. I will post that script here for anyone who can pitch in.

Otherwise, let me know if you think my experience seems atypical or unlikely to be encountered by other Caprover users. Also, thanks again to @danielrataj as well as @githubsaturn for your time and efforts. Caprover remains one of my favorite and most frequently used utilities, and it's something I feel very strongly about contributing towards and promoting.

danielrataj:
Hi @nativeit, you are right. I found the same behaviour on my own Caprover installation. Caprover uses Swarm to distribute containers and has an "any" restart policy configured on the service. So it restarts in case of failure or any other issue.
You can override service behaviour under `App Configs` - see [Service Update Override](https://caprover.com/docs/service-update-override.html) to adjust it.
However, I'm unsure if changing the restart policy will be efficient. Maybe changing MaxAttempts from "unlimited" to "3" (example) won't put the server into a coma - I'm an enthusiast, not an expert... or a doctor :)

Another aspect. I noticed some weird logs in the Redis container when it all went down. It might not recover under lower memory conditions (`vm.overcommit_memory` option in `/etc/sysctl.conf` file on the server. Then the Swarm service is trying to get all services up and working and ends up with a server being drop-dead.

There are actually two things I would like to address in terms of Appwrite and struggling with. The _first one_ is the simple one - to be working (and the main issue might be outside of Caprover or even Appwrite but in underlying components with proper server configuration).

The _second one_ is to be able to upgrade Appwrite. Currently is not possible and thus makes it kind of "demo purposes" only.

PR #644, opened 2022-04-11
Made two minor corrections to the existing Percona Server app.

nativeit:

- [n/a] Icon is added as a png file to the logos directory.

---
The relevant changes are fairly self-explanatory from the commit comments, but just to expand on them a bit:

1. While deploying an instance of Percona on my server, I noticed the Docker Hub link for the available tags was pointing to the MariaDB page rather than the relevant Percona tags, so I simply corrected the URL.
2. Removed the redundant service name from the user-defined app name. 

Quick question on this point: Am I missing a good reason for naming things like this, or is it just a differing preference? I have encountered this particular way of labeling services so many times now that I am not sure if it's intentional, or if it's because folks aren't aware that just `$$cap_appname` by itself is acceptable (or maybe others don't dwell on such small matters), but the `$$cap_appname` in the original file was `$cap_appname-percona`, so when I initially deployed the app, having named my instance "percona", what got deployed was `percona-percona.domain.tld`, which is unnecessary and frequently requires a convoluted process to correct. 

My position is that if one wants the service's name appended onto the user-defined name, they can simply type it into the app name field manually, there's no reason for this to be hard-coded into the template. That said--@githubsaturn if you have a specific preference one way or the other, I will adhere to it moving forward.

### DIFF
--- a/public/v4/apps/percona.yml
+++ b/public/v4/apps/percona.yml
@@ -17,7 +17,7 @@ caproverOneClickApp:
         - id: $$cap_percona_version
           label: Percona Version
           defaultValue: 'ps-8.0'
-          description: Check out their Docker page for the valid tags https://hub.docker.com/_/mariadb?tab=tags
+          description: Check out their Docker page for the valid tags https://hub.docker.com/_/percona
           validRegex: /^([^\s^\/])+$/
         - id: $$cap_db_pass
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks for the PR!

Regarding the naming, no preference on my end.

PR #645, opened 2022-04-13
Update Mastodon

christopher-kapic:

### DIFF
--- a/public/v4/apps/mastodon.yml
+++ b/public/v4/apps/mastodon.yml
@@ -1,39 +1,5 @@
 captainVersion: 4
 services:
-    # Mastodon
-    $$cap_appname:
-        environment:
-            LOCAL_DOMAIN: $$cap_appname.$$cap_root_domain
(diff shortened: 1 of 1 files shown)


christopher-kapic:
I was having problems deploying Mastodon. Since the main container takes a while to deploy, it would timeout before the other containers got a chance to be deployed. Therefore, if we move Mastodon to the end, the other containers will deploy, and then Mastodon will deploy. If it times out, it is possible that the container will finish deploying and everything will work, even though it throws an error on the frontend.

Note: This should probably be tested in a fresh CapRover environment. I'm not sure if the Docker cache had anything to do with my test results.

githubsaturn:
The ordering of items in the file has nothing to do with the ordering of deployment. If such order is needed, you need to make use of `depends_on` flag, example:
https://github.com/caprover/one-click-apps/blob/d4edd8530d7e843857871934c0a42d89da6d0e21/public/v4/apps/nextcloud.yml#L22-L25

christopher-kapic:
Oh, thank you. I didn't know that.

PR #653, opened 2022-04-24
Update FileRun to latest version [broken image, fails to deploy]

d0x7:
Old image no longer exists (therefore this one click app fails to deploy) as it has been moved to FileRun's own DockerHub account and LibreOffice isn't an option anymore. I am using the latest tag, for two reasons: 1) it's the same as it has been before already 2) there are no tags besides latest and arm32/64.

### DIFF
--- a/public/v4/apps/filerun.yml
+++ b/public/v4/apps/filerun.yml
@@ -1,7 +1,7 @@
 captainVersion: 4
 services:
     $$cap_appname:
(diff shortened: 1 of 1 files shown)


githubsaturn:
That's unfortunate that the developers of some applications can't follow simple guidelines to have proper versioning. 

Thanks for the PR.

PR #656, opened 2022-04-29
Add Mayan EDMS

borjarobert:
First working draft of Mayan EDMS as a One-Click-App

Needs an icon, proper instructions and probably some fine tuning, but it works as expected. I'm opening the PR in case anyone wants to work on it too.


borjarobert:
I think it is ready to ship

public/v4/apps/mayan-edms.yml
```diff
@@ -0,0 +1,158 @@
...
+                - CMD exec redis-server --appendonly 'no' --databases '3' --maxmemory 100mb --maxclients '500' --maxmemory-policy allkeys-lru --save '' --requirepass $$cap_redis_pass
+            notExposeAsWebApp: 'true'
+
+    #PostgreSQL
+    $$cap_appname-db:
+        image: postgres:$$cap_postgres_version
+        volumes:
+            - $$cap_appname-db-data:/var/lib/postgresql/data
+        restart: always
+        command:
```
githubsaturn, line 47: `command` is ignored by CapRover. If it's needed convert to dockerfileLines, otherwise, let's remove it

public/v4/apps/mayan-edms.yml
```diff
@@ -0,0 +1,158 @@
...
+            Mayan EDMS is an opensource web-based Document Management System. 
+            Most settings can be set inside the app, but environment variables will override them.
+            Minimum requirements 2GB RAM and 2vCPUs. Recommended 4GB RAM and 4vCPUs
+        end: >-
+            Done!
+            Your service is available at http://$$cap_appname.$$cap_root_domain
+    displayName: 'Mayan EDMS'
+    isOfficial: false
+    description: Mayan EDMS is an open source web-based Document Management System with many advanced features.
+    documentation: https://www.mayan-edms.com/
```
githubsaturn, line 158: Suggest putting this website in instructions.start as well as it makes it easier for people to understand what this is.

borjarobert: Done. Good idea

public/v4/apps/mayan-edms.yml
borjarobert: I was trying to be faithful to the recommended docker-compose.yml but it works fine with the postgres defaults, so I'll leave it like that and don't use DockerfileLines. Thanks

githubsaturn:
Thanks!

PR #658, opened 2022-05-01
Add CouchDB with Clouseau (Full Text Search) as one click app

mqtik:
CouchDB has search natively, however, this can't be enabled without Clouseau - Lucene. To do this, we need a new one click app, that has two containers, one for couch, and another one for Lucene.

- [✅ ] I have tested the template using the method described in README.md thoroughly
- [ ✅] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [ ✅] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [✅ ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [⚠️ ] Icon is added as a png file to the logos directory —the CouchDB logo is in /logos directory, however, I don't want to create a new image like couchdb-clouseau with the same image, what should I do about this?


githubsaturn:
Yes, please add a separate image. All apps, even though similar need to have the same png file with the same name.

githubsaturn:
Thanks!

PR #677, opened 2022-07-06
Update outline to v0.64+ & add OIDC connection & add SMTP config & doc

matbgn:


githubsaturn:
Looks like a complex setup! Thank you!

somnathrakshit:
Is it mandatory to have S3, OIDC, and SMTP keys? Can the app run without these keys?

matbgn:
Unfortunately, Outline is a great product, but which obfuscate its installation to keep other paid solutions on the top. So actually it's not possible to bypass those points.

PR #685, opened 2022-07-24
Updates Wordpress to v6 and MySQL to v8

brahyam:
Updates Wordpress App default versions to:
- Wordpress from v4.9 (giving outdated PHP version warning) to v6.0.1
- MySQL from v5.7 to v8.0.29


githubsaturn:
Thanks!

PR #693, opened 2022-08-06
Added Baserow app

maietta:

- [✓] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.

### DIFF
--- a/public/v4/apps/elasticsearch.yml
+++ b/public/v4/apps/elasticsearch.yml
@@ -6,47 +6,35 @@ services:
             - $$cap_appname-elasticsearch-data:/usr/share/elasticsearch/data
         restart: always
         environment:
-            ES_JAVA_OPTS: -Xms512m -Xmx512m
+            CLI_JAVA_OPTS: -Xms512m -Xmx512m
             cluster.name: $$cap_elasticsearch_cluster_name
-            cluster.initial_master_nodes: $$cap_elasticsearch_cluster_initial_master_nodes
+            discovery.type: $$cap_elasticsearch_discovery_type
             http.port: $$cap_container_port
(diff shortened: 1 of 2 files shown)


maietta:
I ran the `npm run formatter-write` command after seeing this "All checks have failed", but it prettier'ed all the apps. I chose not to commit that and instead now assuming I can ignore this step as it's not outlined in the Readme.

githubsaturn:
It's probably because you've modified package-lock file. Do this:
- Revert `package-lock.json` file to the original
- Run `rm -rf ./node_modules && npm ci`
- Run `npm run formatter-write`

nativeit:
Quick note on this, the Baserow public URL needs to contain http/https or the container frontend will present the user with an error. Obviously, if this is hard-coded into the ENV in the template as http it will break if/when https is enabled (and vise versa). I would think the best approach would be to code it as "https" and include the instructions to enable https in the post-deployment notes.

While there are minor changes happening along these lines, I noted that this monorepo container takes a while to finish loading, as it's a single container package that combines several fairly resource-intensive services (redis, celery, redbeat) so it may be worthwhile to include a warning that it may take a few minutes to become available, and maybe even a note in the start instructions that 2GB/2c minimum is recommended. Lastly, a link to the Docker Hub repo and instructions would be helpful, there's also a massive number of ENV variables available, they're listed at https://gitlab.com/bramw/baserow/-/blob/develop/docker-compose.yml, including that could be useful, but it's nothing that will make the difference in functionality.

Thanks for putting this up, I am really looking forward to having this on my app server. Feel free to let me know if there's anything I can do to help!

PR #704, opened 2022-08-20
Add airflow one click app

Bass-03:


public/v4/apps/airflow.yml
```diff
@@ -0,0 +1,218 @@
...
+            AIRFLOW_EMAIL: $$cap_airflow_email
+            AIRFLOW__SMTP__SMTP_HOST: $$cap_smtp_host
+            AIRFLOW__SMTP__SMTP_PORT: $$cap_smtp_port
+            AIRFLOW__SMTP__SMTP_STARTTLS: $$cap_smtp_starttls
+            AIRFLOW__SMTP__SMTP_SSL: $$cap_smtp_ssl
+            AIRFLOW__SMTP__SMTP_USER: $$cap_smtp_user
+            AIRFLOW__SMTP__SMTP_PASSWORD: $$cap_smtp_password
+            AIRFLOW__SMTP__SMTP_MAIL_FROM: $$cap_smtp_mail_from
+        ports:
+            - '8080:8080'
```
githubsaturn, line 166: Why does it need a mapped port if it's already added as an HTTP container port?

githubsaturn (commented):
Thanks for the PR! Just one minor comment.

public/v4/apps/airflow.yml
Bass-03: hey @githubsaturn 
That is the way the suggested docker compose file is set, I assumed I needed to define that port for that service, then tell caprover to map it to http and https ports.

My assumption is those two things are independent and both need to be explicitly defined.
Isn't that the case?

githubsaturn: Yes, when you've added the port as the container port, the mapping is typically not needed anymore. I suggest removing it and do a quick test to make sure it works.

Bass-03:
hey @githubsaturn 

You were right, it is not needed, I removed the ports and tested, all works as expected.
I checked and the image itself defaults to ports 8080 anyway, so all good.

githubsaturn:
Thanks!

PR #707, opened 2022-09-01
[WIP] add pi hole

LiquidITGuy:

- [] I have made sure that instructions.start and instructions.end are clear and self-explanatory.


LiquidITGuy:
the deploy works but my docker knowledge with network is too bad to make it works (I can only configure the dns on the local server) 
if someone want to continue the pull request to make pihole fully functionnal it will be great

itxtoledo:
@LiquidITGuy what you need to make this script work? I can help

LiquidITGuy:
@itxtoledo it seems to work but when I tried to use pihole as dns server I'm stucked.
I'm not realy sure about how can I unlock it.

michelbarnich:
> @itxtoledo it seems to work but when I tried to use pihole as dns server I'm stucked. I'm not realy sure about how can I unlock it.

What exactly do you mean? Are you setting the Server as your only DNS sever on your local machine? Where does it get stuck?

PR #708, opened 2022-09-04
Fix Listmonk one-click app (wiping DB every time)

MYerouPaleBlue:
Currently, every time the container is restarted the DB is wiped-out! 

This is due to `idempotent = false` by default (doc: "make --install run only if the database isn't already setup"). Setting ` --idempotent` ensures that DB will be initialized only if empty (i.e. the first time).

[source 1](https://github.com/knadh/listmonk/blob/57dbb9e5dbf11b40ad97bbdfe3c39bb9119baea7/cmd/init.go#L108), [source 2](https://github.com/knadh/listmonk/blob/e99c8ed86b71bbcf8bfb2e5361147cec7137cf75/cmd/install.go#L26), [source 3](https://github.com/knadh/listmonk/blob/e99c8ed86b71bbcf8bfb2e5361147cec7137cf75/cmd/install.go#L46)

### DIFF
--- a/public/v4/apps/listmonk.yml
+++ b/public/v4/apps/listmonk.yml
@@ -27,7 +27,7 @@ services:
             containerHttpPort: 9000
             dockerfileLines:
                 - FROM listmonk/listmonk:$$cap_listmonk_version
-                - CMD yes | ./listmonk --install && ./listmonk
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #710, opened 2022-09-06
Bump Elasticsearch and Kibana to v8

Niek:
This PR bumps the one-click-apps of Elasticsearch and Kibana to v8.

Since ES v8 authentication is done in a completely different way than ES7, I decided to disable the security features by default and remind the user to enable HTTP Basic Auth. The alternative is much harder for a new setup:
* Download the auto-generated SSL certificate from the running container
* Scan the logs to the auto-generated password, or execute a command to change it after bootstrap
* Connect to the instance using the downloaded self-signed certificate and the password

I personally think that disabled auth + basic auth is the best way to set up a quick ES8 cluster. See also e.g.: https://github.com/cars10/elasticvue/wiki/Elasticsearch-8

### DIFF
--- a/public/v4/apps/elasticsearch.yml
+++ b/public/v4/apps/elasticsearch.yml
@@ -6,47 +6,35 @@ services:
             - $$cap_appname-elasticsearch-data:/usr/share/elasticsearch/data
         restart: always
         environment:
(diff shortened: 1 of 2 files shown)


githubsaturn:
Thanks!

PR #715, opened 2022-09-11
FEAT: add gotify

LiquidITGuy:

Capture d’écran 2022-09-11 à 20 26 32


public/v4/apps/gotify.yml
```diff
@@ -0,0 +1,26 @@
+captainVersion: 4
+services:
+  '$$cap_appname':
+    image: gotify/server:$$cap_version
+    volumes:
+      - '$$cap_appname-data:/app/data'
+      - '$$cap_appname-config:/etc/gotify/config.yml'
```
githubsaturn, line 7: I don't think I've seen any volume names being mapped to a file. Typically they are mapped to the directory. Does this work just fine? I assume you've tested this?

matbgn: Double this warning as all my tests with direct file map failed.

LiquidITGuy: in https://gotify.net/docs/config we can see 
>gotify/server looks in the following paths for config files
./config.yml
/etc/gotify/config.yml

This file (even if not documented in docker file) is required if you want to customize your gotify

matbgn: It's not about the config file need but its syntax.

Should it not be something like that instead:

- '$$cap_appname-config:/etc/gotify/'

Did you try your template with a fresh install on Caprover?

LiquidITGuy: I just tried to redeploy with the same file on my pr and it seems to work.
I will try to change to follow the request as requested but I was thinking docker can mount a file as a volume if only the file need to be keeped.
Capture d’écran 2022-11-30 à 16 24 54

LiquidITGuy: updated and tried ready @matbgn :)

public/v4/apps/gotify.yml
```diff
@@ -0,0 +1,26 @@
+captainVersion: 4
+services:
+    '$$cap_appname':
+        image: gotify/server:$$cap_version
+        volumes:
+            - '$$cap_appname-data:/app/data'
+            - '$$cap_appname-config:/etc/gotify/'
+caproverOneClickApp:
+    variables:
+        - id: '$$cap_version'
```
matbgn, line 10: As really tiny advice I will more prefer a variable called cap_gotify_version but anyway seems to work, so... cool!!

public/v4/apps/gotify.yml
matbgn: Good job! I only have a tiny a best practice advice (see below) for maintenance purpose only but other than that well done!

public/v4/apps/gotify.yml
LiquidITGuy: thanks for the advice, I will do it next time :)

githubsaturn:
Thanks!

PR #721, opened 2022-09-23
improved photoprism one-click template

vbbot:
_The previous pull request got closed because I changed the branch name for better organisation._

Made the following changes to the template:
1. Added mariadb as the default database driver
2. Added the environment variable to set the default admin password
3. Added a persistent volume for /photoprism/storage which allows settings, cache, databases, etc to persist across container restarts
4. Added an option for running a scheduler to automatically index files in the originals folder using ofelia
5. General improvements to the template instructions, etc.
6. Some other minor tweaks too minor to be included in this list (which I also conveniently can't remember)

PS: I am not prefixing "cap" at the start of the variables because I feel the user defined variables get confused with the system-defined variables such as $$cap_appname or $$cap_root_domain, etc

### DIFF
--- a/public/v4/apps/photoprism.yml
+++ b/public/v4/apps/photoprism.yml
@@ -48,6 +48,7 @@ services:
     $$cap_appname-ofelia:
         restart: unless-stopped
         image: "mcuadros/ofelia:$$ofelia_version"
+        command: daemon --docker
         volumes:
             - "/var/run/docker.sock:/var/run/docker.sock:ro"
         caproverExtra:
--- a/public/v4/apps/photoprism.yml
+++ b/public/v4/apps/photoprism.yml
@@ -1,58 +1,52 @@
 captainVersion: 4
 services:
     $$cap_appname:
-        image: "photoprism/photoprism:$$app_version"
+        image: 'photoprism/photoprism:$$app_version'
         depends_on:
             - $$cap_appname-db
         environment:
             PHOTOPRISM_ADMIN_PASSWORD: $$app_admin_password
             PHOTOPRISM_AUTH_MODE: password
-            PHOTOPRISM_SITE_URL: "https://$$cap_appname.$$cap_root_domain"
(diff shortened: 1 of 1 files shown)


public/v4/apps/photoprism.yml
```diff
@@ -1,23 +1,119 @@
...
+            MARIADB_DATABASE: photoprism
+            MARIADB_USER: photoprism
+            MARIADB_PASSWORD: $$db_password
+            MARIADB_ROOT_PASSWORD: $$db_root_password
+        caproverExtra:
+            notExposeAsWebApp: "true"
+    $$cap_appname-ofelia:
+        restart: unless-stopped
+        image: "mcuadros/ofelia:$$ofelia_version"
+        command: daemon --docker
```
githubsaturn, line 51: Same as the other command... it's an unsupported property.

public/v4/apps/photoprism.yml
```diff
@@ -1,23 +1,119 @@
...
+            $$cap_appname-ofelia.enabled: $$ofelia_enabled
+            $$cap_appname-ofelia.job-exec.photoprism_index.schedule: "$$ofelia_index_schedule"
+            $$cap_appname-ofelia.job-exec.photoprism_index.command: "photoprism index --cleanup"
         caproverExtra:
-            containerHttpPort: '2342'
+            containerHttpPort: "2342"
+    $$cap_appname-db:
+        restart: unless-stopped
+        image: "mariadb:$$db_version"
+        command: >-
```
githubsaturn, line 32: This is an unsupported property. If it's important, use a different way to override the command. See this for example:
https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/redis.yml#L10-L12

vbbot: Cool, shall make the changes in another commit!

githubsaturn:
Thanks!

PR #725, opened 2022-09-25
added baikal app template

vbbot:

Alright, this is the last pull request for Baikal. Pretty sure I got this right this time haha. Thanks for bearing with me!


public/v4/apps/baikal.yml
```diff
@@ -0,0 +1,26 @@
...
+        volumes:
+            - '$$cap_appname-data:/var/www/baikal/Specific'
+            - '$$cap_appname-config:/var/www/baikal/config'
+        caproverExtra:
+            containerHttpPort: '80'
+caproverOneClickApp:
+    variables:
+        - id: '$$cap_version'
+          label: Baikal Version
+          defaultValue: 'nginx'
```
githubsaturn, line 14: This is a mutable tag. Meaning that if you deploy today, you might be installing 0.9.2-nginx, if you deploy tomorrow, you might be installing 0.9.3-nginx. 

We should always use a fixed tag, in this case I'd change this to `0.9.2-nginx`

public/v4/logos/baikal.png
```diff
@@ -0,0 +1,1477 @@
+
+
+
+
+
+
+
+<!DOCTYPE html>
```
githubsaturn, line 8: This is not a PNG file, it's an html file!

I assume https://sabre.io/img/baikal.png is the right PNG file?

public/v4/apps/baikal.yml
```diff
@@ -0,0 +1,26 @@
...
+          description: Check out their Docker page for the valid tags https://hub.docker.com/r/ckulka/baikal/tags
+          validRegex: "/^([^\\s^\\/])+$/"
+    instructions:
+        start: |-
+            Baïkal is a lightweight CalDAV+CardDAV server. It offers an extensive web interface with easy management of users, address books and calendars. It is fast and simple to install and only needs a basic php capable server. The data can be stored in a MySQL or a SQLite database.
+             More details: https://sabre.io/baikal/
+        end: |-
+            Baïkal  has been successfully deployed!
+    displayName: Baïkal (vbbot)
```
githubsaturn, line 23: `(vbbot)` should be removed - unless the name of the app is actually `Baïkal (vbbot)` ?

githubsaturn (commented):
Thanks! I know the process isn't very easy... so thanks for keeping up with it!

The PR looks good. Just a few minor comments.

Let me know if you're having difficulties making these edits. In that case, I'll be more than happy to take over and apply these changes myself.

public/v4/logos/baikal.png:Zone.Identifier
```diff
@@ -0,0 +1,4 @@
+[ZoneTransfer]
```
githubsaturn, line 1: Added by mistake I assume? :) 
I'll merge this as is and will remove this file myself.

Issue #482, opened 2021-08-23
App Suggestion: Excalidraw


jluterek:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
Excalidraw - https://github.com/excalidraw/excalidraw

**Do you know if there is any official Docker image for the app?**

We publish a Docker image with the Excalidraw client at [excalidraw/excalidraw](https://hub.docker.com/r/excalidraw/excalidraw). You can use it to self-host your own client under your own domain, on Kubernetes, AWS ECS, etc.

```sh
docker build -t excalidraw/excalidraw .
docker run --rm -dit --name excalidraw -p 5000:80 excalidraw/excalidraw:latest
```

matbgn:
I'm working on it

PR #726, opened 2022-09-25
Excalidraw

matbgn:


matbgn:
Closes #482

public/v4/apps/excalidraw.yml
```diff
@@ -0,0 +1,77 @@
...
+    $$cap_appname-storage-backend:
+        image: kiliandeca/excalidraw-storage-backend:$$cap_storage_backend_version
+        restart: always
+        environment:
+            STORAGE_URI: redis://srv-captain--$$cap_appname-redis:6379
+            STORAGE_TTL: 2592000000
+        depends_on:
+            - $$cap_appname-redis
+        ports:
+            - '8080:8080'
```
githubsaturn, line 29: I assume this is not needed as the containerHttpPort is already set to 8080.

public/v4/apps/excalidraw.yml
```diff
@@ -0,0 +1,77 @@
...
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+        volumes:
+            - $$cap_appname-redis-data:/data
+
+caproverOneClickApp:
+    variables:
+        - id: $$cap_excalidraw_version
+          label: Excalidraw version
+          defaultValue: 'latest'
```
githubsaturn, line 44: image

public/v4/apps/excalidraw.yml
```diff
@@ -0,0 +1,77 @@
...
+          description: Check out this Docker page for any valid tags https://hub.docker.com/r/kiliandeca/excalidraw/tags
+        - id: $$cap_storage_backend_version
+          label: Excalidraw storage backend adapter version
+          defaultValue: 'latest'
```
githubsaturn, line 48: Same `latest`

githubsaturn (commented):
Thanks! Just a few minor comments.

public/v4/apps/excalidraw.yml
matbgn: Yep leaving it

public/v4/apps/excalidraw.yml
matbgn: Actually [kiliandeca/excalidraw](https://hub.docker.com/r/kiliandeca/excalidraw/tags) is only providing latest tag **but I already mapped it with a variable** so it can be easily adapted by end user. Other suggestion?

public/v4/apps/excalidraw.yml
matbgn: Dito

public/v4/apps/excalidraw.yml
githubsaturn: ah this really sucks :/ it's the dumbest way of using Docker... :man_facepalming: 

1- I suggest you open an issue with the maintainer of that Docker image and have them fix this issue.

2- As a temporary fix, you can do something like this:
https://github.com/caprover/one-click-apps/blob/48d2870e6e5769acce66a4a14712968f347683bf/public/v4/apps/mongodb-backup-s3.yml#L21

matbgn: 1. I already was dreaming of doing it... It seems that it was only a dream after all :thinking: Here right now from dream to action: https://gitlab.com/kiliandeca/excalidraw-fork/-/issues/2

matbgn: 2. """Nice""" workaround... even if it's really not a good solution to work like this as you said.

githubsaturn:
Thanks!

PR #744, opened 2022-10-01
Removing appsmith no database

aqidd:

This version of appsmith deployment is no longer compatible for version 1.8.0 and above. This was made because the early docker deployment still need some manual things to do. With the introduction of the official docker image (added https://github.com/caprover/one-click-apps/pull/594) then this version is better removed


githubsaturn:
Thanks!

PR #756, opened 2022-10-09
feat: add gotenberg

ronaldloyko:
This pull request adds the one-click-app for [Gotenberg](https://gotenberg.dev).

## Checks


public/v4/apps/gotenberg.yml
```diff
@@ -0,0 +1,310 @@
...
+                    --webhook-disable=$WEBHOOK_DISABLE
+                    --prometheus-collect-interval=$PROMETHEUS_COLLECT_INTERVAL
+                    --prometheus-namespace=$PROMETHEUS_NAMESPACE
+                    --prometheus-disable-collect=$PROMETHEUS_DISABLE_COLLECT
+                    --prometheus-disable-route-logging=$PROMETHEUS_DISABLE_ROUTE_LOGGING
+                    --log-format=$LOG_FORMAT
+                    --log-level=$LOG_LEVEL
+                    --gotenberg-graceful-shutdown-duration=$GRACEFUL_SHUTDOWN_DURATION
+        environment:
+            VERSION: $$cap_GOTENBERG_VERSION
```
githubsaturn, line 86: I don't think these environment variables (also `ARG`s in the dockerfileLines) are needed as your just passing the variables directly to the command line... I might be wrong, but maybe test it out without these and see if it works.

Merging anyways. But if it works without these, let's open another PR to clean it up. Thanks!

ronaldloyko: I added the environment variables to be able to edit them later after deployment.

Regarding the `ARG`s in the `dockerfileLines` you are absolutely right, no need to define `ARG`s as we can just use `${var}` directly.

Thanks for the hints.

Since this PR is still open, I can edit it right here if you like.

ronaldloyko: Ok, just tested.

When I remove the `ARG`s and I want to change an environment variable later after deployment, the change is not reflected since it just restarts the container with the environment variables from first creation during one-click-app deployment.

My approach with `ARG`s will recreate the container with the changed environment variables.
Therefore, I suggest accepting the PR as it is now, unless there are any other objections.

githubsaturn: Ah I see! Makes sense! Thanks for confirming that!

PR #762, opened 2022-10-15
 iRedMail - Update requirements

ido-pluto:
I am adding basic requirements for iRedMail that appear before the installation.


githubsaturn:
Thanks!

PR #767, opened 2022-10-16
chore(n8n): bump version and optimize variables

ronaldloyko:
This pull request updates the one-click-application for [n8n](https://n8n.io).

## Changes

- Bumped versions to latest stable releases
  - n8n: `0.198.2`
  - PostgreSQL: `15`
- Removed unused Docker socket volume
- Removed environment variables for basic authentication since n8n provides a user account system
- Removed unnecessary environment variables like `VUE_APP_URL_BASE_API`, `WEBHOOK_TUNNEL_URL`
- Added environment variable to opt-out of telemetry
- Added environment variable for application environment

## Checks

### DIFF
--- a/public/v4/apps/n8n-io.yml
+++ b/public/v4/apps/n8n-io.yml
@@ -27,6 +27,7 @@ services:
             N8N_HOST: $$cap_appname.$$cap_root_domain
             N8N_DIAGNOSTICS_ENABLED: $$cap_N8N_DIAGNOSTICS_ENABLED
             WEBHOOK_URL: https://$$cap_appname.$$cap_root_domain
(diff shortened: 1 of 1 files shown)


LiquidITGuy:
It seems works on my server thanks :) 
WIth this version, SMTP server is mandatory to enable other account, should they be added to the configuration env var ? (or at least add a link to this for the end explaining that first and say to go quick on the page to prevent any steal of the main account ?)

ronaldloyko:
After deployment, the first account created will be the owner/admin account. So I think there is no chance to steal it and this is also fine for sole users.
But nevertheless you are absolutely right when it comes to inviting other members and for instances used by teams. 
I will add the environment variables for optional SMTP configuration.
Thanks for the hint.

githubsaturn:
Thanks!

PR #768, opened 2022-10-17
Postgres updates

zenflow:
A few little fixes and enhancements to the postgres one-click app.
Please see each commit message for descriptions of the changes.

Regarding the last commit (a2ccd72531794f78f9bc19bf941cd57c6dbe0a9f):
The current behavior (of post-fixing the user-given app name with "-db") is super annoying...
I would like to be able to specify an app name to be used unaltered.
I can choose an appropriate app name myself.
If I forget that this one-click app does this post-fixing, my app name ends up being something like "my-app-db-db", and I have to delete it and recreate it 

### DIFF
--- a/public/v4/apps/postgres.yml
+++ b/public/v4/apps/postgres.yml
@@ -16,7 +16,7 @@ caproverOneClickApp:
     variables:
         - id: $$cap_postgres_version
           label: Postgres Version
-          defaultValue: '14.1'
+          defaultValue: '14.5'
           description: Check out their Docker page for the valid tags https://hub.docker.com/r/library/postgres/tags/
           validRegex: /^([^\s^\/])+$/
         - id: $$cap_pg_user
--- a/public/v4/apps/postgres.yml
+++ b/public/v4/apps/postgres.yml
@@ -1,9 +1,9 @@
 captainVersion: 4
 services:
-    $$cap_appname-db:
+    $$cap_appname:
         image: postgres:$$cap_postgres_version
         volumes:
-            - $$cap_appname-db-data:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


public/v4/apps/postgres.yml
```diff
@@ -42,7 +44,7 @@ caproverOneClickApp:
             After installation on CapRover, it will be available as srv-captain--YOUR_CONTAINER_NAME at port 5432 to other CapRover apps.
 
             Enter your Postgres Configuration parameters and click on next. It will take about a minute for the process to finish.
-        end: "Postgres is deployed and available as srv-captain--$$cap_appname-db:5432 to other apps. For example with NodeJS: 'const client = new Client({ user: 'cap_pg_user', host: 'srv-captain--$$cap_appname', database: 'cap_pg_db', password: '********', port: 5432})'"
+        end: "Postgres is deployed and available as srv-captain--$$cap_appname:5432 to other apps. For example with NodeJS: 'const client = new Client({ user: '$$cap_pg_user', host: 'srv-captain--$$cap_appname', database: '$$cap_pg_db', password: '********', port: 5432})'"
```
ronaldloyko, line 47: I think we can also show the password here instead of `********`.
What do you think?

zenflow: Absolutely. The `********` makes no sense since the password is displayed in plaintext in the environment variables section anyways (which is great!).

githubsaturn: Good suggestion! It's particularly okay because it's supposed to be the initial password and the user is expected to change post initialization for better security. In fact, after the installation is done, the password can be removed from the env vars and the app works just fine!

Feel free to open a new PR to update this. But for now, merging as is.

PR #778, opened 2022-10-29
add FreshRSS

lcx:
Add FreshRSS from official docker image

> FreshRSS is a self-hosted RSS and Atom feed aggregator.
> It is lightweight, easy to work with, powerful, and customizable.

Since I know we are all lazy with backups, I have included an automated PSQL Dump to create backups. Unfortunately this docker image only has a "latest" tag.


githubsaturn:
Backup app is an interesting idea. Wouldn't it be better if it's a standalone one click app? Users will be able to use it for all apps that use Postgres under the hood, not just FreshRSS.

christopher-kapic:
Postgres backups would be a great feature for Caprover in general. I have a [half-baked repo](https://github.com/christopher-kapic/pg_capup), but an official feature would be nice.

lcx:
ok, removed the backup from freshrss. The issue is that people like me are lazy, they will postpone backups for some other day and at some point forget completely that's why I prefer to add the backup in the one click to prevent people forgetting. But maybe something that can be used for all containers would make more sense.

githubsaturn:
Thanks!

PR #780, opened 2022-11-01
Add ntfy one click app

nico-kow:

- [ ☑️ ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [ ☑️ ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [ ☑️ ] Icon is added as a png file to the logos directory.


public/v4/apps/ntfy.yml
```diff
@@ -0,0 +1,77 @@
+# Use schema version 4
+captainVersion: 4
+
+
+# One-click-app options
+caproverOneClickApp:
+
+    # App name
+    displayName: ntfy.sh (nico-kow)
```
githubsaturn, line 9: Please remove the username from the display name

public/v4/apps/ntfy.yml
```diff
@@ -0,0 +1,77 @@
...
+    # Description displayed on the app list
+    description: |-
+        Send push notifications to your phone or desktop via PUT/POST
+
+    # Define instructions
+    instructions:
+
+        # Displayed one the app config page, where you enter all variable values
+        start: |-
+            ![ntfy Logo](https://github.com/binwiederhier/ntfy/raw/main/web/public/static/img/ntfy.png)
```
githubsaturn, line 20: "main" is mutable, use a fixed commit for your icon.

githubsaturn:
Thanks!

PR #798, opened 2022-12-02
chore(name) add all missing displayName

LiquidITGuy:

### DIFF
--- a/public/v4/apps/droppy.yml
+++ b/public/v4/apps/droppy.yml
@@ -20,6 +20,6 @@ caproverOneClickApp:
             droppy is deployed and available as $$cap_appname.
 
             IMPORTANT: It will take up to 2 minutes for droppy to be ready. Before that, you might see 502 error page.
-    displayName: 'droppy [archived]'
+    displayName: 'droppy'
     description: droppy is a self-hosted file storage server with a web interface
     documentation: Read the documentation @ https://github.com/droppyjs/droppy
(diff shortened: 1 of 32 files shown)


LiquidITGuy:
I add all missing or empty displayName by checking (with a script I've made) the missing one. Then for all of them I check on the official website to write in the same way the display name.

public/v4/apps/droppy.yml
```diff
@@ -20,6 +20,6 @@ caproverOneClickApp:
             droppy is deployed and available as $$cap_appname.
 
             IMPORTANT: It will take up to 2 minutes for droppy to be ready. Before that, you might see 502 error page.
-    displayName: ''
+    displayName: 'droppy [archived]'
```
githubsaturn, line 23: archived?

LiquidITGuy: Droppy has been [archived](https://github.com/silverwind/droppy) and is not maintain anymore since few years 
It use node 12 and may introduce some security failure.
As the owner explain itself that the project is archived 
>Development on droppy has ceased because I don't have enough time or motivation to properly support it and because of its outdated technology stack, it became exceedingly boring to work.

I add this tag but I would be happy to write it anywhere if there is a way to specify this information in any part of the yml file

LiquidITGuy: (in the about we can read : **ARCHIVED** Self-hosted file storage)

githubsaturn: This seems to be a different droppy then? https://github.com/droppyjs/droppy

LiquidITGuy: You are totally righ I'll fix that right now

githubsaturn:
Thanks!

PR #799, opened 2022-12-04
Feature/vaultwarden upgrade vaultwarden

LiquidITGuy:


arthuragone:
Since 1.30.0 NGIX congig has to change. 

As stated in official GitHub Doc: 

> Note: The WebSockets service for live sync has been integrated in the main HTTP server, which means simpler proxy setups that don't require a separate rule to redirect WS traffic to port 3012. Please check the updated examples in the [wiki](https://github.com/dani-garcia/vaultwarden/wiki/Proxy-examples). It's recommended to migrate to this new setup as using the old server on port 3012 is deprecated, won't receive new features and will be removed in a future release.

Maybe someone could make an update to the oneclick config? (On my side I can't make it work but neither previous version)

adamshand:
We already have BitwardenRS which is the old name for Vaultwarden (but the recipe is current).  Perhaps we can delete this and just rename BitwardenRS to Vaultwarden?

PR #805, opened 2022-12-09
add Greenlight to one-click-apps

andrekutianski:
Greenlight 2.0 (referred hereafter as simply “Greenlight”) is a Ruby on Rails application that provides a simple interface for users to create rooms, start meetings, and manage recordings.

overview: https://docs.bigbluebutton.org/greenlight/gl-overview.html


githubsaturn:
Thanks!

PR #815, opened 2022-12-21
remove regex for smtp username

### LINKED ISSUE #814: App issue: plausible - wrong smtp user regex
The plausible installer forces a regex for usernames and returns an error for the username if it doesn't match the regex: 

`Invalid value. Does not match Regex:/^([^\s^\/])+$/`

this should be left to the user, trying to find a regex matching whatever the RFC defines as valid username can be hard. 
Im using self hosted postal (https://github.com/postalserver/postal) to send my notification mails which creates users like 'foo/bar' which are valid but are rejected by this regex.

lcx:
closes #814

didn't update the rest as I just fixed the issue with regex and didn't want to update / change anything else. 


githubsaturn:
Thanks!

PR #822, opened 2022-12-27
Update Fider App

aqidd:
- Update fider env
- Using latest fider as default


githubsaturn:
Thanks!

lexoyo:
Hi @aqidd 
Great to see fider here :)) (i discovered it because it is here actually ;)
Would you add the database like it is done with other apps?
Or maybe you could give me instructions on how to add the DB by hand?

aqidd:
Hi @lexoyo , to add DB manually just add this env variable

DATABASE_URL=<postgres-connection-string>

lexoyo:
That makes sense thank you

PR #825, opened 2023-01-02
ARM64 version

pasmer:

I've just created a copy of Rallly one-click-app using ARM64 docker image link.
Thank you for your attention
PM

### DIFF
--- a/public/v4/apps/rallly_arm64.yml
+++ b/public/v4/apps/rallly_arm64.yml
@@ -1,0 +1,90 @@
+captainVersion: 4
+services:
+    $$cap_appname:
+        restart: always
+        depends_on:
+            - $$cap_appname-postgres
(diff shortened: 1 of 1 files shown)


githubsaturn:
If everything is the same across the two except the image name, instead of duplicating everything, just extract out the image name as a variable and have the default value to point to the regular version and add the new image name in the description of the variable:
```
...........
        image: '$$cap_rally_image:$$cap_rallly_version'
...
...
...
...
        - id: $$cap_rally_image
          label: Rallly image name
          defaultValue: lukevella/rallly
          description: >-
              change to "justsky/rally" if installing on an arm64 device
          validRegex: '/^(lukevella\/rallly|justsky\/rally)$/'





```

githubsaturn:
feel free to open another PR!

PR #829, opened 2023-01-04
Improve outline deployment

douglara:


public/v4/apps/outline.yml
```diff
@@ -64,6 +64,10 @@ services:
             SMTP_TLS_CIPHERS: $$cap_smtp_tls_ciphers
             SMTP_SECURE: $$cap_smtp_secure
             DEFAULT_LANGUAGE: $$cap_default_language
+        caproverExtra:
+            - FROM outlinewiki/outline:$$cap_outline_version
+            - EXPOSE 3000
+            - CMD yarn db:migrate --env production-ssl-disabled; yarn start
```
githubsaturn, line 70: Is this tested? It looks like the command here is different from the snippet below?  `db:migrate` vs `sequelize:migrate` - are these both the same?

iamsid: https://github.com/outline/outline/issues/4772
It seems sequelize is no longer available.

githubsaturn:
@iamsid - just to confirm your comment on the other thread, are you saying that even this PR does not work?

Should we just remove outline from the directory until it's fixed by someone?

iamsid:
Did a bit more digging to identify the issue. And was able to install Outline.
Turns out, this merge-request/PR is not needed. But there are two issues with the existing YAML file.

Unfortunately, I don't know how to do a PR. So here are the two issues:
1. The pre-deploy script in current YAML file does not show correctly in the UI after installation is complete. Its marked using ``` this YAML code, but it shows incorrectly in CapRover's UI. The pre-deploy script is needed for the app to run the migrations and start working.
2. The 4th point in this section, called "Run DB migration", needs to be "Paste below pre-deploy script in App Config tab". And the pre-deploy script itself just needs to change `sequelize:migrate` to `db:migrate`.

public/v4/apps/outline.yml
douglara: Maybe `sequelize:migrate` is deprecated command, I followed error instruction to fix

douglara:
Sorry I created this PR with error

douglara:
ixe

douglara:
Sorry, I created this PR with error

Follow fixes 

@iamsid 
1 - I configured to run migration on application startup, now no needs step 4
2 - Now runs automatically

3 - Removed optional environment variables raising errors if not configured
4 - Add instruction to configure authentication method, outline works only thirdy party authentication methods

githubsaturn:
Thanks!

Issue #808, opened 2022-12-12
App Suggestion: [OpenSpeedTest]


openspeedtest:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
OpenSpeedTest is a Self-Hosted Network SpeedTest Application. 
[![Download OpenSpeedTest-Server V2.1](https://github.com/openspeedtest/v2-Test/raw/main/images/10G-S.gif)](https://github.com/openspeedtest/Speed-Test#self-hosted-on-premise--docker-imagesource-code  "Docker OpenSpeedTest-Server")

**Do you know if there is any official Docker image for the app?**
Official Docker Image https://hub.docker.com/r/openspeedtest/latest
Default port for http : 3000
Default port for http : 3001

````
sudo docker run --restart=unless-stopped --name openspeedtest -d -p 3000:3000 -p 3001:3001 openspeedtest/latest
````

PR #834, opened 2023-01-10
added openspeedtest

Svelte-Enthusiast:
Added  #808 , openspeedtest


public/v4/apps/openspeedtest.yml
```diff
@@ -0,0 +1,24 @@
...
+
+caproverOneClickApp:
+    variables:
+        - id: '$$cap_openspeedtest_version'
+          label: OpenSpeedTest Version
+          defaultValue: 'speedtest'
+          description: Check out their Docker page for the valid tags https://hub.docker.com/r/openspeedtest/latest/tags
+    instructions:
+        start: |-
+            OpenSpeedTest is a self-hosted speedtest.
```
githubsaturn, line 18: Add a note linking to their website. Something like:
```
OpenSpeedTest is a self-hosted speedtest. You can read more about this on https://openspeedtest.com/selfhosted-speedtest
```

public/v4/apps/openspeedtest.yml
```diff
@@ -0,0 +1,24 @@
...
+        end: |-
+            OpenSpeedTest has been successfully deployed! You must enable https for the app to work. Go to the app setting and enable https.
+    displayName: OpenSpeedTest
+    isOfficial: true
+    description: |-
+        OpenSpeedTest is a self-hosted speedtest written in NodeJS.
```
githubsaturn, line 24: I don't see any nodejs in their docs, can you link where you found this?

openspeedtest:
@Svelte-Enthusiast Thank you for the PR.
 
#### Default port for http : 3000
#### Default port for https : 3001

- Written in Vanilla Javascript (Client-Side). No PHP, Node Js, or any other Server-Side programming languages are used.  
- This is a docker implementation using nginxinc/nginx-unprivileged:stable-alpine. uses significantly fewer resources.
- Nginx Deliver our static SpeedTest application, That is how it works.

I don't know how caprover handles apps. 
If you run behind a reverse proxy and reverse proxy is used for handling HTTPS.
I think it's better to run the app in HTTP mode, to save server resources. 
3000 is the HTTP port.
-  If you run it behind a reverse proxy you should increase the [post-body content length](https://github.com/openspeedtest/Speed-Test/issues/4#issuecomment-1229157193) to 35 Megabytes or more.

githubsaturn:
Thanks! Formatting seems to be not working. Suggest running `npm run formatter-write` to reformat your changes.

public/v4/apps/openspeedtest.yml
```diff
@@ -2,23 +2,23 @@ captainVersion: 4
 
 services:
     '$$cap_appname':
-        image: index.docker.io/openspeedtest/latest:$$cap_openspeedtest_version
+        image: index.docker.io/test/latest:$$cap_test_version
```
githubsaturn, line 5: lots of wrong replacement here. All "openspeedtest" instances in the definition are converted to test

public/v4/apps/openspeedtest.yml
```diff
@@ -2,23 +2,23 @@ captainVersion: 4
...
-          label: OpenSpeedTest Version
+        - id: '$$cap_test_version'
+          label: Test Version
           defaultValue: 'speedtest'
-          description: Check out their Docker page for the valid tags https://hub.docker.com/r/openspeedtest/latest/tags
+          description: Check out their Docker page for the valid tags https://hub.docker.com/r/test/latest/tags
     instructions:
         start: |-
-            OpenSpeedTest is a self-hosted speedtest. You can read more about this on https://openspeedtest.com/selfhosted-speedtest.
+            Test is a self-hosted speedtest. You can read more about this on https://test.com/selfhosted-speedtest.
```
githubsaturn, line 18: same here... **Test** is a self-hosted speedtest. You can read more about this on https://**test.com**/selfhosted-speedtest.

Svelte-Enthusiast: super weird. the only thing i did was run the formatter. i fixed it with the next push

githubsaturn:
Thanks!

@openspeedtest  - I assume you maintain this image. I suggest you change the image to a standard format.

`openspeedtest/latest:openspeedtest` is a non conventional format. Docker images should have their version as a tag. So for example: `openspeedtest/openspeedtest:v1.2.3` or `openspeedtest/openspeedtest:latest`

openspeedtest:
@githubsaturn When I started using docker, I was unaware of or fully understood this concept. 
Now I don't want to break existing deployments by deleting them. Also, it will reset the pull count and star count.
No breaking changes will be introduced to this image, especially to the latest and speedtest tags. That is why I avoided versioning them. Still, if that is important from the next version (2.5.5), I will add a version number.
Thanks!

githubsaturn:
Yea I definitely recommend using concrete version tags. Here is a snippet on [why they are important](https://vsupalov.com/docker-latest-tag/#:~:text=You%20should%20avoid%20using%20the,apart%20from%20the%20image%20ID.).

nativeit:
I was getting ready to open a PR for this app, I may link it here if I can confirm that I'm not breaking anything, or that I wasn't making a mistake in my initial setup. The problem as it currently exists is this:

When deployed as-is, the port is set to the `https` port `3001`, but whether or not SSL is enabled under the app's HTTP settings, this results in the following Nginx error:

`The plain HTTP request was sent to HTTPS port`

It's my understanding that this is due to the way the reverse proxy is being handled. I'm not an expert with Nginx, so a second opinion would be good here, but I believe the app's port should be set to the `http` port `3000` instead, as Caprover will handle the secure connection, and the traffic to port `3000` will effectively be local proxy traffic.

I deployed `v2.0.5` to my server, and with the HTTP port set to `3001` it produces the error with/without HTTPS enabled, but with `3000` set it works as expected with either HTTP/HTTPS enabled in the app's settings.

My PR was going to change the HTTP port to `3000`, and the default version tag to `v2.0.5`. I'd welcome @openspeedtest to please advise, as well as anyone with better knowledge of the Nginx proxy to confirm that using `3000` with SSL enabled under the app settings achieves the same security as running OpenSpeedTest from a non-proxied web server on port `3001`.

nativeit:
While reading on the `Content-Length` header issue posted above, I also saw that there were issues with `http2`/`http3`, and under the repo's Nginx configuration ([link](https://github.com/openspeedtest/Nginx-Configuration/blob/main/OpenSpeedTest-Server.conf)) it includes a block disabling `http2`.

Caprover's default Nginx config uses `listen 443 ssl http2;`

I don't think it's generally ideal for apps to require editing the default config (although it isn't unheard of). That said, I noticed if "Websocket Support" has been enabled, then this line is added to the default config: `proxy_http_version 1.1;` 

@githubsaturn - Would the "Websocket Support" flag do anything to help with this, or is it only triggered when the `$http_upgrade` header is included with the client's request (otherwise all requests get passed to `443 ssl http2` as usual)?

It seems like running OpenSpeedTest behind a reverse proxy may just have some inherent issues. I am running it under port `3000` in a Caprover instance (Proxmox KVM) that's hosted on a bare metal server at OVH's Virginia datacenter (my nearest region) and it seems to be working normally, and providing realistic measurements for my current location. 

I'll be glad to continue testing, if the `http2` vs `http1.1` topic is ultimately not relevant to the Caprover deployment, I'll move this particular issue to the OpenSpeedTest repo.

nativeit:
I attempted to work with both the `http/1.1` vs `http/2` and `request_max_body_size` issues, but in the end could not determine whether any of my changes to the Nginx configuration made an impact at all (assuming they didn't break something outright, which also happened in a few cases). No matter what I did, when I used my browser's Network Inspection it was _always_ `http/2`, but I'm also fairly sure the Docker networking layer facilitates those connections. Correct me if I'm wrong, but in order to see the relevant stats I would need to inspect the traffic between the upstream server and reverse proxy. 

Tldr; I don't have any suggestions with regard to the reverse proxy problems. But I believe I have fixed the port issue, and I added details in the deployment instructions regarding the limits for this implementation of OpenSpeedTest. I put my local branch up at https://github.com/nativeit/one-click-apps/tree/openspeedtest-dev - if you'd like me to open a PR, I certainly will.

githubsaturn:
> but I believe the app's port should be set to the http port 3000 instead, as Caprover will handle the secure connection

100% correct!

RE: the rest of your questions, I can't say much since it's really dependent of how this app was built.
Looks like they have some guide here for reverse proxy:
https://github.com/openspeedtest/Docker-Image?tab=readme-ov-file#or-use-docker-composeyml

But looks like they require a lot of custom settings for nginx to work properly 
https://github.com/openspeedtest/Nginx-Configuration/blob/main/OpenSpeedTest-Server.conf

PR #838, opened 2023-01-19
Update openvscode-server.yml

redwerkz:
Bump to latest version 1.74.3
Additional note about characters allowed for `token`

- [x] I have tested the template


githubsaturn:
Thanks!

PR #840, opened 2023-01-21
update portainer to 2.16.2

ivnnv:
This exact template still works for the current latest version

I can use itself to check it is indeed using this very version 😄 

Screenshot 2023-01-21 at 13 22 02


githubsaturn:
Thanks!

PR #841, opened 2023-01-21
Update passbolt to version 3

ariaieboy:
I've changed the volume mount point according to the passbolt version 3.

And I've made the DB version configurable instead of being fixed to a specific version


githubsaturn:
Thanks!

PR #845, opened 2023-01-23
Add Mumble

Vdauphin:


githubsaturn:
Thanks for the PR! Please run `npm run formatter-write` to reformat your file.

Vdauphin:
I am confused, I rerun `npm run formatter-write` but git doesn't show me any change and the error message from checks doesn't help me to catch what is wrong in the file

githubsaturn:
Weird :/ 

Anyways, I've re-ran this for you and it's all good now.

PR #862, opened 2023-02-07
Add Audiobookshelf

beejsbj:


public/v4/apps/audiobookshelf.yml
```diff
@@ -0,0 +1,37 @@
+captainVersion: 4
+services:
+  $$cap_appname:
+    image: ghcr.io/advplyr/audiobookshelf:$$cap_version
+    ports:
+      - 13378:80
```
githubsaturn, line 6: Why do you need port mapping here?

public/v4/apps/audiobookshelf.yml
```diff
@@ -0,0 +1,37 @@
...
+  variables:
+    - id: "$$cap_version"
+      label: Audiobookshelf Version
+      defaultValue: "2.2.14"
+      description: Check out their Docker page for the valid tags https://hub.docker.com/r/advplyr/audiobookshelf/tags
+      validRegex: "/.{1,}/"
+
+    - id: $$cap_path_to_audiobookshelf
+      label: Location of app folder in host
+      description: path to base folder that will contain your audiobooks, podcasts, metadata and config.
```
githubsaturn, line 24: This is not clear to the end user. If this directory doesn't exist, the app is going to crash on start up. If any of the subdirectories don't exist, again, the app will crash.

Instead, I suggest just use volume name, and in instructions.end ask the user to remap the directories to where they have their content.

public/v4/apps/audiobookshelf.yml
```diff
@@ -0,0 +1,37 @@
...
+      validRegex: /\/[^\s-][a-zA-Z0-9_\/-]*[^\/][^\s-]/
+
+  instructions:
+    start: |-
+      Audiobookshelf is a self-hosted audiobook and podcast server. Official Docker image.
+    end: |-
+      Audiobookshelf has been succesfully deployed!
+      It is available at http://$$cap_appname.$$cap_root_domain or at 123.123.123.123:13378
```
githubsaturn, line 32: Again, why port mapping is needed? Wouldn't the regular domain be just fine?

githubsaturn (commented):
Thanks for the PR! Just a few minor comments.

public/v4/apps/audiobookshelf.yml
beejsbj: their docs perform this mapping, I thought they might have some internal config set to it.

https://www.[audiobookshelf](https://www.audiobookshelf.org/docs/#docker-install).org/docs/#docker-install

public/v4/apps/audiobookshelf.yml
```diff
@@ -19,9 +19,19 @@ caproverOneClickApp:
       description: Check out their Docker page for the valid tags https://hub.docker.com/r/advplyr/audiobookshelf/tags
       validRegex: "/.{1,}/"
 
-    - id: $$cap_path_to_audiobookshelf
-      label: Location of app folder in host
-      description: path to base folder that will contain your audiobooks, podcasts, metadata and config.
+    - id: $$cap_path_to_audiobooks
+      label: Location of audiobooks directory
+      description: full path to the directory where the audiobooks are stored
```
githubsaturn, line 24: Make it explicit that this path MUST EXISTS on the host, otherwise the app won't start.

Same with the other paths.

public/v4/apps/audiobookshelf.yml
githubsaturn: This needs to be removed. The port mapping is only needed because the default config is not using reverse proxy.

githubsaturn (commented):
Thanks! One last thing: the build is failing, you need to look into it. Perhaps you need to run `npm run formatter-write`

public/v4/apps/audiobookshelf.yml
```diff
@@ -0,0 +1,45 @@
...
+          label: Location of metadata directory
+          description: full path to the directory where the metadata is stored. Path on host MUST exist for the app to start!
+          validRegex: /\/[^\s-][a-zA-Z0-9_\/-]*[^\/][^\s-]/
+
+    instructions:
+        start: |-
+            Audiobookshelf is a self-hosted audiobook and podcast server. Official Docker image.
+        end: |-
+            Audiobookshelf has been succesfully deployed!
+            It is available at http://$$cap_appname.$$cap_root_domain or at 123.123.123.123:13378
```
githubsaturn, line 40: `123.123.123.123:13378` still needs to be removed. I'll remove it after merge.

githubsaturn (commented):
Thanks!

PR #864, opened 2023-02-11
Update changedetection.yml

redwerkz:
- [x] I have tested the template and have a running and functional instance


githubsaturn:
Thanks!

PR #871, opened 2023-02-24
Add breadcord app

alf1e:


githubsaturn:
Thanks for the PR. Looks like the image is not a popular image. Only been download 2 times. 

We only limit the main repo to popular apps (thousands of stars and/or tens of thousands of downloads). Otherwise, we'll end up with a one click app repo that has thousands of app and will be impossible to manager.

Having said that, you are more than welcome to create your own 3rd party one click app repository. CapRover is built in a way that it's got full support for 3rd party one click app repository exactly for these use cases.

image

PR #875, opened 2023-03-01
Add Firefly III Data Importer (FIDI)

marcusique:


public/v4/apps/fireflyiii-data-importer.yml
```diff
@@ -0,0 +1,51 @@
...
+        image: fireflyiii/data-importer:$$cap_fidi_version
+        environment:
+            FIREFLY_III_ACCESS_TOKEN: $$cap_ff_token
+            FIREFLY_III_URL: $$cap_ff_url
+            NORDIGEN_ID: $$cap_nordigen_id
+            NORDIGEN_KEY: $$cap_nordigen_key
+            SPECTRE_APP_ID: $$cap_spectre_app_id
+            SPECTRE_SECRET: $$cap_spectre_secret
+        ports:
+            - $$cap_fidi_port:8080
```
githubsaturn, line 13: You probably don't need this. This is normally needed when you don't have reverse proxy. But with CapRover, your app is available via  `appname.root.domain.com`

public/v4/apps/fireflyiii-data-importer.yml
```diff
@@ -0,0 +1,51 @@
+captainVersion: 4
+services:
+    $$cap_appname:
+        image: fireflyiii/data-importer:$$cap_fidi_version
+        environment:
+            FIREFLY_III_ACCESS_TOKEN: $$cap_ff_token
+            FIREFLY_III_URL: $$cap_ff_url
```
githubsaturn, line 7: Do you need `$$cap_ff_url` as a separate variable? Can you not replace it with `$$cap_appname.$$cap_root_domain` ?

public/v4/apps/fireflyiii-data-importer.yml
marcusique: You're probably right. I'm running my installation locally and ran into port issues in the past. I'll fix this.

public/v4/apps/fireflyiii-data-importer.yml
marcusique: This is not the data importer URL, but a Firefly III URL (which is a separate app). It can either live on local machine or outside, hence it's needed as a separate variable.

githubsaturn:
Thanks!

Issue #878, opened 2023-03-04
App Suggestion: serpbear


jonas-hoebenreich:
**Please tell us what app you'd like to see on CapRover as a one-click app.**
SerpBear is an Open Source Search Engine Position Tracking App. It allows you to track your website's keyword positions in Google and get notified of their positions.
https://github.com/towfiqi/serpbear

**Do you know if there is any official Docker image for the app?**
https://hub.docker.com/r/towfiqi/serpbear

uhlhosting:
@jonas-hoebenreich have you managed to get serpbear running with GSC, and if so how you made the variables to use the. GCP private key  in one line?

PR #879, opened 2023-03-04
add serpbear

jonas-hoebenreich:
see https://github.com/caprover/one-click-apps/issues/878

https://github.com/towfiqi/serpbear
https://hub.docker.com/r/towfiqi/serpbear


public/v4/apps/serpbear.yml
```diff
@@ -0,0 +1,59 @@
+captainVersion: 4
+services:
+    $$cap_appname:
+        restart: always
+        image: 'towfiqi/serpbear:$$cap_serpbear_version'
+        environment:
+            NEXT_PUBLIC_BASE_URL: 'http://$$cap_appname.$$cap_root_domain'
```
githubsaturn, line 7: Where is `NEXT_PUBLIC_BASE_URL` being used?

githubsaturn: Ah I think you meant to use `NEXT_PUBLIC_APP_URL`

public/v4/apps/serpbear.yml
```diff
@@ -0,0 +1,59 @@
...
+            PASSWORD: $$cap_serpbear_password
+            SECRET: $$cap_serpbear_secret
+            APIKEY: $$cap_serpbear_apikey
+            SESSION_DURATION: $$cap_serpbear_session
+        caproverExtra:
+            containerHttpPort: '3000'
+        volumes:
+            - '$$cap_appname-serpbear-data:/app/data/serpbear'
+        networks:
+            - serpbear-net
```
githubsaturn, line 18: This is not used by CapRover, you can remove.

public/v4/apps/serpbear.yml
jonas-hoebenreich: Good catch, thanks for noticing

public/v4/apps/serpbear.yml
jonas-hoebenreich: ok 👌

githubsaturn:
Thanks!

uhlhosting:
@githubsaturn first of all this is not official supported app, since the person who submitted it is not the owner. Some very important feature of the app does not seem to work. And even the owner does not seem to find a solution to it. Yet adding new apps should not get the official app badge, unless submitted and maintained by their makers. Hope this makes sense.

githubsaturn:
> first of all this is not official supported app

@uhlhosting  - where do you get this? As far as I can see, it's the image that is linked in the official repository. What makes you think that it's not the official image?

uhlhosting:
@githubsaturn my bad, the image uses the official app indeed, yet I tough official apps are the ones submitted and maintained by their makers.

PR #897, opened 2023-04-05
Update PeerTube to v5

Vdauphin:

### DIFF
--- a/public/v4/apps/peertube.yml
+++ b/public/v4/apps/peertube.yml
@@ -44,6 +44,7 @@ services:
             PEERTUBE_SMTP_TLS: $$cap_smtp_tls
             PEERTUBE_SMTP_DISABLE_STARTTLS: $$cap_smtp_starttls
             PEERTUBE_ADMIN_EMAIL: $$cap_admin_email
+            PT_INITIAL_ROOT_PASSWORD: $$cap_admin_password
             PEERTUBE_SECRET: $$cap_secret
         caproverExtra:
             containerHttpPort: '9000'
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

davidleebooth:
Hello, I seem to be directed here and hopefully I am in the right spot.

I am using the 1 click install for PeerTube. I started pre version 5.x and was proceeding to upgrade via docker image name.
As per upgrading instructions on peertube, I generated a base64 secret for the env vars that is now required, but additionally in PeerTube 5.x there are some additional NGINX requirements involving websockets and websocket plugin locations. 

[PeerTube standard NGINX](https://github.com/Chocobozzz/PeerTube/blob/develop/support/nginx/peertube)
(lines 115-146 are websocket specific.)

I had initially thought I did have them entered into caprover's NGINX correctly and have enabled websocket support, but after the upgrade this error remained:

360.e3d42fe8da2a6e0a.js:1 WebSocket connection to 'wss://davbot.media/socket.io/?accessToken=84c48cf643b74457935dd4b81d7dc2ffbf8f6ddd&EIO=4&transport=websocket&sid=PlliITpNBJ6W2uiqAAAU' failed

Also, when live streaming, Checking on federated servers with that stream, the stream does not play and a websocket error occuring there as well, I am unsure if these two things are related but it could be possible.

I tried putting these statements into NGINX in different ways, even replacing proxy pass in the PT lines with $upstream as its written in the caprover default lines for proxy pass, and going so far as to paste the whole thing over just to see what would happen with no change.

I will admit that I am not the most familiar with NGINX at this time, and the issue is likely PEBKAC, but any help and pointers would be greatly appreciated.

Thanks.

PR #899, opened 2023-04-07
Add KeyDB

l0rem:
Hi, i would love to add [KeyDB](https://docs.keydb.dev/) to one-click-apps. 

It's pretty popular alternative (fork) to redis, at moment of writing it has 7.2k stars on [GH](https://github.com/Snapchat/KeyDB), 1M+ pulls on [DH](https://hub.docker.com/r/eqalpha/keydb) and some pretty big companies use it :)

Thanks!


l0rem:
Maybe its a good idea to include prettier in the check-list, just to avoid a bit of confusion? 🙂

githubsaturn:
Thanks!

geoman2:
Hi, did you try DragonflyDB, is supposed to be up to x25 more performant than redis, but can't prove. What about KeyDB, first time heard about it

l0rem:
@geoman2 i would suggest reading this thread - https://news.ycombinator.com/item?id=31796311

I have pretty much no experience with anything other than redis or keydb, so for now i am sticking with what works for me, but you should not be pressured into using what others are telling you to use. build your next thing with dragonfly and share your experiences, it would be interesting for me and many others :)

PR #900, opened 2023-04-08
Adding checkmk to One-Click-Apps

drewOP:
Added checkmk raw edition to One-Click-Apps
https://checkmk.com/

- ☑️ I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter`


public/v4/apps/checkmk.yml
```diff
@@ -0,0 +1,33 @@
+captainVersion: 4
+services:
+    '$$cap_appname':
+        image: checkmk/check-mk-raw:$$cap_version
+        environment:
+            TZ: '$$cap_tz'
+            PHP_TZ: '$$cap_tz'
```
githubsaturn, line 7: I don't see these two env vars in checkmk's docs. Also the associated CapRover variable doesn't exist in the variables section.

public/v4/apps/checkmk.yml
```diff
@@ -0,0 +1,33 @@
...
+        volumes:
+            - '$$cap_appname-data:/omd/sites'
+            - /etc/localtime:/etc/localtime:ro
```
githubsaturn, line 10: `:ro` is not read by CapRover. Can be removed.

public/v4/apps/checkmk.yml
drewOP: just some copy and paste errors. Fixed it.

public/v4/apps/checkmk.yml
drewOP: removed it

githubsaturn:
Thanks!

PR #905, opened 2023-04-10
Use Awes0meHub repository as exemple in Readme

Vdauphin:
[Awes0meHub](https://github.com/Awes0meHub/caprover-one-click-apps) repository is the second most active repository for CapRover One-Click-Apps
I suggest to use it as exemple but also as an alternative for apps which doesn't fit the scope of this repository


githubsaturn:
Thanks!

PR #909, opened 2023-04-15
Update GlitchTip to v3.1.2

sr258:
I've updated GlitchTip and its dependencies to the latest versions.

- [X] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter`

These self checks don't apply for upgrades:


githubsaturn:
Thanks!

PR #927, opened 2023-05-13
fix(chatwoot): Missing host to link to! Please provide the :host parameter

minkbear:

When I install one-click chatwoot and play, I got the error `ActionView::Template::Error (Missing host to link to! Please provide the :host parameter, set default_url_options[:host],`. So the solution is add environment variable `FRONTEND_URL` to get it works as https://github.com/chatwoot/chatwoot/issues/5707#issuecomment-1287271013

- [ ☑️] I have tested the template using the method described in README.md thoroughly
- [ ☑️] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [ ☑️] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [ ☑️] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [ ☑️] Icon is added as a png file to the logos directory.
- [ ☑️] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter`
- [ ☑️] I will take resposibility addressing any issues that arises as a result of this PR (maintaining this app).

### DIFF
--- a/public/v4/apps/chatwoot.yml
+++ b/public/v4/apps/chatwoot.yml
@@ -48,6 +48,7 @@ services:
         environment:
             RAILS_ENV: production
             RAILS_LOG_TO_STDOUT: 'true'
+            FRONTEND_URL: http://$$cap_appname.$$cap_root_domain
             SECRET_KEY_BASE: $$cap_chatwoot_secret_key_base
             POSTGRES_HOST: srv-captain--$$cap_appname-postgres
             POSTGRES_DATABASE: chatwoot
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

uhlhosting:
@minkbear can you bring this up to date now according to: https://www.chatwoot.com/blog/v2-17 we need following changes:

- Redis 7.x from 6.x
- and a library libvips, yet is not clear to me why they did not included the library in the docker image themselves...

minkbear:
> @minkbear can you bring this up to date now according to: https://www.chatwoot.com/blog/v2-17 we need following changes:
> 
> * Redis 7.x from 6.x
> * and a library libvips, yet is not clear to me why they did not included the library in the docker image themselves...

there is libvips in docker. see https://github.com/chatwoot/chatwoot/commit/8b3d13a9208d84c5f9073b5abb047306779b6ace so this will be in [v2.17.1](https://hub.docker.com/layers/chatwoot/chatwoot/v2.17.1/images/sha256-285a80f99c897c68cff6e8a7486a31a4423ce1683afc33d9f3f113a8e6f0769b?context=explore)

for upgrade redis version please see https://caprover.com/docs/one-click-apps.html#upgrading-one-click-apps

uhlhosting:
@minkbear that upgrade procesure for redis does not works. It will upgrade redis instance alone, not the dump . that will be remaining on 6.2.6

PR #940, opened 2023-05-27
Update pocketbase.yml to work on ARM

erickythierry:
- I made some simple changes to the template so that it is possible to choose the cpu architecture when installing 
- I also modified the volume mount point so that it is possible to update the pocketbase binary without having to delete and recreate the app
- I also updated the default installation version to the most current one (0.16.3)


githubsaturn:
Thanks!

PR #947, opened 2023-06-06
Update windmill.yml to latest version and added support for docker socket

slig:
HI all, since they added Docker support and AFAIK it's not possible to change a running service to add persistence, I've updated the spec. Also, removed defaults from the `environment` vars.

### DIFF
--- a/public/v4/apps/windmill.yml
+++ b/public/v4/apps/windmill.yml
@@ -19,28 +19,25 @@ services:
         environment:
             DATABASE_URL: postgres://postgres:$$cap_postgres_pass@srv-captain--$$cap_appname-postgres/windmill?sslmode=disable
             BASE_URL: $$cap_wm_base_url
-            RUST_LOG: 'info'
-            NUM_WORKERS: '0'
-            DISABLE_SERVER: 'false'
-            METRICS_ADDR: 'false'
+            RUST_LOG: info
+            NUM_WORKERS: 0
         depends_on:
             - $$cap_appname-postgres
         caproverExtra:
-            containerHttpPort: '8000'
+            containerHttpPort: 8000
 
     $$cap_appname-worker:
         image: ghcr.io/windmill-labs/windmill:$$cap_app_version
+        volumes:
+            - $$cap_appname-worker-dependency-cache:/tmp/windmill/cache
+            - /var/run/docker.sock:/var/run/docker.sock
         environment:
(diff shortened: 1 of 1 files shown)


public/v4/apps/windmill.yml
```diff
@@ -84,12 +81,11 @@ caproverOneClickApp:
     variables:
         - id: $$cap_app_version
           label: Windmill Version
-          defaultValue: '1.99'
+          defaultValue: '1.109.1'
           description: Checkout their github page for the valid tags https://github.com/windmill-labs/windmill/releases
           validRegex: /^([^\s^\/])+$/
         - id: $$cap_wm_base_url
           label: WM Base URL
-          defaultValue: https://$$cap_appname.$$cap_root_domain
```
githubsaturn, line 92: Is this removed intentionally?

public/v4/apps/windmill.yml
```diff
@@ -19,28 +19,25 @@ services:
...
         depends_on:
             - $$cap_appname-postgres
         caproverExtra:
-            containerHttpPort: '8000'
+            containerHttpPort: 8000
 
     $$cap_appname-worker:
         image: ghcr.io/windmill-labs/windmill:$$cap_app_version
+        volumes:
+            - $$cap_appname-worker-dependency-cache:/tmp/windmill/cache
```
githubsaturn, line 32: I'm not familiar with Windmill, but is making the cache persistent really necessary? Usually caches are supposed to be transient as the name suggests....

githubsaturn (commented):
Thanks! Just some minor questions.

public/v4/apps/windmill.yml
slig: Yes. I thought that it would be replaced in the setup screen, but as far as I tested, it still appears as the literal string `https://$$cap_appname.$$cap_root_domain`.

public/v4/apps/windmill.yml
slig: I'm following the original `docker-file`. In the previous one-click version, before having the `docker socket`, I did not pay attention to this. But now with the docker socket, we do have the need of persistence and the caching is supposed to speed up all the packages installs, so I added it.

slig:
Thanks for reviewing, @githubsaturn ! I followed up with the Windmill author and he approved this new one-click definition yesterday.

githubsaturn:
Perfect, thanks!

PR #948, opened 2023-06-06
Add Yourls

steigerdich:
First pull request... :)

After I had problems setting up the URL shortener Kutt I made a YAML for YOURLS.


public/v4/apps/yourls.yml
```diff
@@ -0,0 +1,82 @@
...
+            MYSQL_USER: $$cap_db_user
+            MYSQL_PASSWORD: $$cap_db_pass
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+    $$cap_appname:
+        depends_on:
+            - $$cap_appname-db
+        image: yourls:$$cap_yourls_version
+        volumes:
+            - $$cap_appname-logs:/usr/src/yourls/logs
```
githubsaturn, line 20: Not familiar with yourls - but is mapping the logs as a persistent volume really necessary?

public/v4/apps/yourls.yml
```diff
@@ -0,0 +1,82 @@
...
+    instructions:
+        start: >-
+            YOURLS is a set of PHP scripts that will allow you to run Your Own URL Shortener.
+        end: >
+            yourls is ready for final steps at $$cap_YOURLS_SITE/admin .
+            In case you have set another Instance URL than http://$$cap_appname.$$cap_root_domain: Don't forget to connect! (and enable HTTPS)
+
+            Have fun!
+    displayName: yourls
+    isOfficial: false
```
githubsaturn, line 80: I think you're using the official image.

githubsaturn (commented):
Thanks! 

Just a few minor comments/questions. Also, a few checks from the self checkbox list are missing.

public/v4/apps/yourls.yml
steigerdich: You are right, not necessary!

public/v4/apps/yourls.yml
steigerdich: True! Changed...

steigerdich:
I will make a few more tests tomorrow, maybe it would be nice to publish another version that already comes with a frontend. I'll check that..

githubsaturn:
Thanks! Can you also please check all the self check list in the PR?

steigerdich:
Small proof that it already works: https://tinyy.de/pryourls ;)

githubsaturn:
Fantastic! Thanks!

PS: no need for a proof, just a self check suffices :smile:

ms4sman:
I may have discovered an issue with this one-click app. The cap_YOURLS_COOKIEKEY doesn't seem to get used anywhere, and possibly other variables. I do not know all the impacts of this, but what I have found is that since the value I supply to that variable is not being used, the Secret Signature Token is always the same in every install I do, which is obviously a security issue. That CookieKey is used to generate said signature token.

Let me know if more information is needed!

Thanks

PR #956, opened 2023-06-28
Update directus-mysql-redis.yml

JonathanDoelan:
Since Directus 10.4.0 we need to use the `REDIS` env var instead of `CACHE_REDIS`. By keeping both vars it will work for previous as well as for future versions.


githubsaturn:
Thanks!

PR #958, opened 2023-06-29
Updated PrivateBin to 1.5.1. Added persistent config volume

ivanskodje:
## Changes
- Updated PrivateBin to version 1.5.1 (see [tags](https://hub.docker.com/r/privatebin/nginx-fpm-alpine/tags))
- Added volume for the config folder to allow overriding PrivateBin configurations. See [configuration](https://github.com/PrivateBin/PrivateBin/wiki/Configuration) for documentation. 

By default the config folder is completely empty, so it wont affect existing or new deployments. It will, however, allow us to add our own `conf.php` based on their template.  

### Help Wanted

If there is a way to edit/add config from within CapRover without needing to SSH into the server and manually create a `conf.php` file, **please let me know**. 

PrivateBin have [configuration file sample](https://raw.githubusercontent.com/PrivateBin/PrivateBin/master/cfg/conf.sample.php) that would be perfect as a can use as a placeholder, with everything that is non-default is commented out. This would be the ideal way to manage custom config. :cat2: 

### DIFF
--- a/public/v4/apps/privatebin.yml
+++ b/public/v4/apps/privatebin.yml
@@ -7,13 +7,14 @@ services:
             PHP_TZ: '$$cap_tz'
         volumes:
             - '$$cap_appname-data:/srv/data'
+            - '$$cap_appname-cfg:/srv/cfg'
         caproverExtra:
             containerHttpPort: '8080'
 caproverOneClickApp:
     variables:
         - id: '$$cap_version'
           label: PrivateBin Version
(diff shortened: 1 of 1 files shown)


githubsaturn:
> If there is a way to edit/add config from within CapRover without needing to SSH into the server and manually create a conf.php file, please let me know.



You can look at the "filebrowser" one click app - it's pretty handy for these situations.

githubsaturn:
Thanks!

PR #981, opened 2023-08-23
Update sshd.yml, enables ssh tunneling by defaul

slig:
Updates a ENV var so that the openssh-server enables by default the ssh tunneling.

The docs says that the SSH one-click install should be the one used for tunneling, but somehow this pinned version ships with the `AllowTcpForwarding  no`  and that makes no sense in this context.

The easy fix is to add this one-liner to the ENV variables:

`DOCKER_MODS=linuxserver/mods:openssh-server-ssh-tunnel`

As suggested here: https://github.com/linuxserver/docker-openssh-server/issues/43#issuecomment-1238950477

Thanks a lot!

### DIFF
--- a/public/v4/apps/sshd.yml
+++ b/public/v4/apps/sshd.yml
@@ -7,6 +7,7 @@ services:
             PGID: 1001
             TZ: Europe/London
             PUBLIC_KEY: $$cap_sshd_public_key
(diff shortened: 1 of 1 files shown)


githubsaturn:
Nice!!!

PR #990, opened 2023-10-03
Data in jsreport now persists container restart

gpistotnik:
Corrected in app volume from /jsreport to /app.

### DIFF
--- a/public/v4/apps/jsreport.yml
+++ b/public/v4/apps/jsreport.yml
@@ -3,7 +3,7 @@ services:
     $$cap_appname:
         image: jsreport/jsreport:$$cap_jsreport_version
         volumes:
-            - $$cap_appname-data:/jsreport
+            - $$cap_appname-data:/app
         restart: always
         environment:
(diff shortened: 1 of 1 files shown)


githubsaturn:
Mounting app directory is `/jsreport` according to their document. Where do you see this new volume?
```
sudo docker run -p 80:5488 -v /myapp:/jsreport jsreport/jsreport:4.0.0
```

githubsaturn:
Also, it may be just for the newer versions. The default version needs to be updated as well.

gpistotnik:
I have version jsreport/jsreport:3.9.0-full and data does not persist if mapped to /jsreport. I have manually checked in container where the report templates are.

gpistotnik:
Mapping /app/data did not work for me, on existing jsreport container /app/data worked but on new installation it didn't, when I changed to /app - according to log file, it did.

githubsaturn:
Interesting! Okay then, let's do this!

githubsaturn:
Can you just update the version to the version you tested with?

PR #992, opened 2023-10-04
update readme and an app to show websocket support

IzzleNizzle:
I created an app with the one-click-app database recently and websocket support was not checked by default. This caused the app to not work appropriately until that was enabled. I posted in slack about this and was encouraged to submit a pull request. Please let me know if any adjustments need to be made. 

I've created a [pull request on the front end repo](https://github.com/caprover/caprover-frontend/pull/132) in order to handle this variable. It will enable the websocket support automatically if `true`. 

Websocket can be checked by adding a property like the following into the template: 

```yml
services:
    $$cap_appname:
        ...

        caproverExtra:
            websocketSupport: 'true'

        ...
```

I tested to ensure this was not a breaking change, and was an opt in only type of feature. From my testing this is friendly to templates with or without the variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ caproverOneClickApp:
     - `dockerfileLines` which is a multiline variable, and can be used instead of `image` property in the service. You must delete the `image` property if you want to use this parameter.
     - `containerHttpPort` is useful when the underlying service uses a custom port for HTTP. If not provided, the default will be `"80"`
     - `notExposeAsWebApp` can be set to `"true"` when the underlying service is not an HTTP app. This is useful for databases and other internally used services.
-    - `websocketSupport` can be set to `"true"` to automatically enable Websocket Support.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,6 +47,7 @@ caproverOneClickApp:
     - `dockerfileLines` which is a multiline variable, and can be used instead of `image` property in the service. You must delete the `image` property if you want to use this parameter.
     - `containerHttpPort` is useful when the underlying service uses a custom port for HTTP. If not provided, the default will be `"80"`
     - `notExposeAsWebApp` can be set to `"true"` when the underlying service is not an HTTP app. This is useful for databases and other internally used services.
+    - `websocketSupport` can be set to `"true"` to automatically enable Websocket Support.
```
githubsaturn, line 50: Add a note saying that it's only supported in versions 1.12+ (future versions)

IzzleNizzle: Good point, I have adjusted the pull request to include that specification.

PR #1007, opened 2023-10-22
Update keila.yml

lexoyo:
You need to **turn on Websocket Support** in the app's HTTP settings so that keila works

If you don't, then you will not have any preview for your campaigns and then they will not be sent at all

- [ ] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter`


githubsaturn:
Thanks!

PR #1012, opened 2023-11-02
feature/homepage

wajeht:


public/v4/apps/homepage.yml
```diff
@@ -0,0 +1,49 @@
...
+          validRegex: /.{1,}/
+        - id: $$cap_homepage_version
+          label: Homepage Version
+          defaultValue: v0.7.4
+          description: Check out their GitHub packages page for the valid tags
+              https://github.com/gethomepage/homepage/releases
+          validRegex: /^([^\\s^\\/])+$/
+        - id: $$cap_config_path
+          label: Configuration Path
+          defaultValue: /path/to/config
```
githubsaturn, line 38: Why don't use a volume name. It's more common and you don't need to make sure that the path exists

package-lock.json
```diff
@@ -1,138 +1,182 @@
 {
   "name": "caprover-one-click-apps",
   "version": "1.0.0",
-  "lockfileVersion": 1,
```
githubsaturn, line 4: Can you please revert the changes in package-lock?

githubsaturn (commented):
Thanks!

public/v4/apps/homepage.yml
wajeht: done! still works. `https://test2.jaw.dev/`

githubsaturn:
Amazing! Thanks!

PR #1013, opened 2023-11-03
Add Official AzuraCast One-Click App Manifest

BusterNeece:
Hello! I'm the creator and lead maintainer of the [AzuraCast](https://github.com/AzuraCast/AzuraCast) web radio management suite. Our application is primarily deployed via Docker, so CapRover makes excellent sense as a deployment method for our users. I'm hoping we can get added to the official app list!


public/v4/apps/azuracast.yml
```diff
@@ -0,0 +1,206 @@
...
+            - '8466:8466'
+            - '8470:8470'
+            - '8475:8475'
+            - '8476:8476'
+            - '8480:8480'
+            - '8485:8485'
+            - '8486:8486'
+            - '8490:8490'
+            - '8495:8495'
+            - '8496:8496'
```
githubsaturn, line 158: Are these ports all required? If so, I suggest adding a warning note in `instructions.start` for the user to make sure all these ports are open and they are not used by any other applications. If they are already in-use the container fail to start up.

githubsaturn:
Thanks for the PR! Just a minor comment. Also the build is failing. Probably some formatting issues.

BusterNeece:
@githubsaturn Thank you! Formatting fixes are applied and I've added a description about the heavy port clearance.

githubsaturn:
Thanks!

PR #1019, opened 2023-11-14
Fix missing volume redis in Nextcloud

Vdauphin:

### DIFF
--- a/public/v4/apps/nextcloud.yml
+++ b/public/v4/apps/nextcloud.yml
@@ -17,6 +17,8 @@ services:
         documentation: Taken from https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/apache/docker-compose.yml
         image: redis:$$cap_redis_version
         restart: always
+        volumes:
+            - $$cap_appname-redis-data:/data
         caproverExtra:
             notExposeAsWebApp: 'true'
     $$cap_appname:


public/v4/apps/nextcloud.yml
```diff
@@ -17,6 +17,8 @@ services:
         documentation: Taken from https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/apache/docker-compose.yml
         image: redis:$$cap_redis_version
         restart: always
+        volumes:
+            - $$cap_appname-redis-data:/data
```
githubsaturn, line 21: redis doesn't have any persistent data by default unless the start command is changed.

Why is this volume needed?

Vdauphin: I don't know exactly but when I checked the volume of the redis docker (with `docker inspect -f '{{ .Mounts }}'`), it has shown a volume with random number (https://stackoverflow.com/questions/62649302/docker-run-volume-keeps-creating-random-volumes-and-not-using-the-one-specifie)
Also it seems to have a volume `/data` define in docker file here: https://github.com/docker-library/redis/blob/master/7.2/alpine3.18/Dockerfile#L103
The CapRover app has a volume also https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/redis.yml#L5

PR #1033, opened 2023-12-02
add wakapi

ariaieboy:
Wakapi is an open-source tool that helps you keep track of the time you have spent coding on different projects in different programming languages and more.
Ideal for statistics freaks and anyone else.
https://wakapi.dev


githubsaturn:
Thanks!

uhlhosting:
@ariaieboy nice job! thanks for this addition, there is only one issue on my deployment, I get a 404 when I try to reach /api . Can you confirm your instance works out of the box with this url?

ariaieboy:
> @ariaieboy nice job! thanks for this addition, there is only one issue on my deployment, I get a 404 when I try to reach `/api` . Can you confirm your instance works out of the box with this url?

Make sure that the wakapi public url env is sets correctly

uhlhosting:
@ariaieboy I set it correctly, since I can visit the domain and all other pages, but the /api .

ariaieboy:
> @ariaieboy I set it correctly, since I can visit the domain and all other pages, but the /api .

that's because there is no /api route go to `yourdomain.com/swagger-ui/swagger-ui/index.html` and check the API endpoints
for example go to this URL `yourdomain.com/api/health` if it returns `app=1
db=1` that means your API working fine

PR #1048, opened 2023-12-29
feat: add dufs

Ysn4Irix:


public/v4/apps/dufs.yml
```diff
@@ -0,0 +1,48 @@
+captainVersion: 4
+
+services:
+  '$$cap_appname':
+    caproverExtra:
+      dockerfileLines:
```
githubsaturn, line 6: https://github.com/sigoden/dufs?tab=readme-ov-file#with-docker

Looks like they have official Docker image for the app. Can we use that instead ?

public/v4/apps/dufs.yml
```diff
@@ -0,0 +1,48 @@
...
+      - Support https
+      - Support webdav
+      - Easy to use with curl
+    end: |-
+      Dufs has been successfully deployed!
+  displayName: Dufs
+  isOfficial: true
+  description: |-
+    Dufs is a distinctive utility file server that supports static serving, uploading, searching, accessing control, webdav...
+    - Serve static files & Download folder as zip file...
```
githubsaturn, line 47: Description should be just one line. No bullet points.

githubsaturn (commented):
Nice app! Just a couple of minor comments. Also the build is failing (probably due to formatting issues)

githubsaturn:
Looks like you accidentally checked in package-lock file. And the build is still failing.

Ysn4Irix:
fixed, i think everything should works fine now 😃

githubsaturn:
Still failing unfortunately. I think you'd have to revert the package-lock changes, then make sure you run `npm run formatter-write` before committing your changes.

githubsaturn:
It still shows a modified package-lock file. Perhaps it's easier to recreate the PR. 
image

PR #1082, opened 2024-03-06
Added FreeScout Helpdesk

houndbee:
* Added freescout.yaml and logofile

- [X ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [X ] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [ X] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [ X] Icon is added as a png file to the logos directory.
- [ X] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- [ X] I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).

### DIFF
--- a/public/v4/apps/freescout.yml
+++ b/public/v4/apps/freescout.yml
@@ -31,7 +31,6 @@ services:
 
 caproverOneClickApp:
     variables:
-
         - id: $$cap_db_user
           label: Database user
           defaultValue: freescout


public/v4/apps/freescout.yml
```diff
@@ -0,0 +1,78 @@
...
+            MYSQL_PASSWORD: $$cap_db_pass
+        caproverExtra:
+            notExposeAsWebApp: 'true'
+
+    $$cap_appname:
+        depends_on:
+            - $$cap_appname-db
+        image: tiredofit/freescout:$$cap_freescout_version
+        volumes:
+            - $$cap_appname-data:/www/html
```
githubsaturn, line 21: Let's remove this as it adds little value and users cannot upgrade simply by deploying new image. This is the directory where the source code lives.

githubsaturn (commented):
Thanks for the PR and apologies for the late response! 

Just one minor comment.

nativeit:
I tried deploying this a while back, and it's functional, but only minimally. To access most common features for a help desk, you must purchase additional modules, and _none_ of the official modules are open source. 

I'm not saying this shouldn't be included in the repo once the open issue is resolved, but to anyone who might be considering using this, it's a point I wish I'd have known before spending time on it, since it touts itself as "free", and open source--it's not, and it isn't.

PR #1089, opened 2024-03-20
Update Immich to v1.99.0

borjarobert:
Just a bump in image versions so it doesn't lag too far behind the mobile app.

Bumped version of immich-server, immich-microservices and immich-machine-learning to latest release (v1.99.0)

Updated pg-vecto-rs database to version 0.2.0

Although this version allows for hardware acceleration, I couldn't figure out a way to make it configurable by users within caprover. It works as-is, but without hardware acceleration

Also, the logo is changed so I changed the png.

Some time ago they also changed the license to AGPLv3, so I updated the description too.


githubsaturn:
Thanks!

githubsaturn:
PS: hardware acceleration isn't straightforward with Docker swarm, example: https://gist.github.com/tomlankhorst/33da3c4b9edbde5c83fc1244f010815c

PR #1106, opened 2024-05-25
Create captain-definition for caprover/one-click-apps

alyssadev:
I have tested this deploying from https://git.aly.pet/aly/caprover-apps to https://caprover-apps.aly.pet/. I plan to add an index page to this third party repo, or add a markdown renderer for README.md, but that's out of scope for this particular pull request and could be added to this repository at a later date. My hope is to make it easier for other users to create third party repositories for CapRover hosted on their own infrastructure.

### ☑️ Self Check before Merge

- [x] I will take responsibility addressing any issues that arise as a result of this PR (maintaining this file).
- ~~I have tested the template using the method described in README.md thoroughly~~
- ~~I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.~~
- ~~I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.~~
- ~~I have made sure that instructions.start and instructions.end are clear and self-explanatory.~~
- ~~Icon is added as a png file to the logos directory.~~
- ~~I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)~~

PR #1111, opened 2024-06-04
Added readme for captain-definition usage

alyssadev:
Per comment on #1106, added readme block for captain-definition


githubsaturn:
Thanks!

PR #1125, opened 2024-07-02
fix: Container HTTP port for Open SpeedTest

benbristow:
Fixing default container HTTP port for Open SeedTest.

Current default (3001) shows a NGINX https over http port error page.


githubsaturn:
Thanks!

PR #1133, opened 2024-07-15
Update sshd app's image to the version-9.7_p1-r4

jorups:
The current version of `sshd.yml` contains very old base image for openssh-server (`lscr.io/linuxserver/openssh-server:version-8.6_p1-r3`).
However, SSH tunnel mod (`linuxserver/mods:openssh-server-ssh-tunnel`) is already ported to the s6 v3 (https://github.com/linuxserver/docker-mods/tree/master?tab=readme-ov-file#new-v3-mods) and the base image doesn't work properly with the updated mod.
More specifically mod does not modify the `sshd_config` file to make SSH tunnel work, hence manual modification of configuration was required.
This should fix the issue https://github.com/caprover/one-click-apps/issues/1041

### DIFF
--- a/public/v4/apps/sshd.yml
+++ b/public/v4/apps/sshd.yml
@@ -1,7 +1,7 @@
 captainVersion: 4
 services:
     $$cap_appname:
-        image: lscr.io/linuxserver/openssh-server:version-8.6_p1-r3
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #1134, opened 2024-07-16
Let user choose GOSTv3 versions

amirsaam:

### DIFF
--- a/public/v4/apps/gogostv3.yml
+++ b/public/v4/apps/gogostv3.yml
@@ -18,10 +18,10 @@ services:
                 -  # Make directories needed
                 - RUN mkdir -p tmpfiles && mkdir -p /configuration
                 -  # Download and prepare GOST binary
-                - RUN wget -q https://github.com/go-gost/gost/releases/download/v3.0.0-nightly.20240711/gost_3.0.0-nightly.20240711_linux_arm64.tar.gz
-                - RUN tar -xzf gost_3.0.0-nightly.20240711_linux_arm64.tar.gz -C tmpfiles && mv tmpfiles/gost /bin/gost && chmod +x /bin/gost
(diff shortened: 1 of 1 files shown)


public/v4/apps/gogostv3.yml
```diff
@@ -18,10 +18,10 @@ services:
                 -  # Make directories needed
                 - RUN mkdir -p tmpfiles && mkdir -p /configuration
                 -  # Download and prepare GOST binary
-                - RUN wget -q https://github.com/go-gost/gost/releases/download/v3.0.0-nightly.20240711/gost_3.0.0-nightly.20240711_linux_arm64.tar.gz
-                - RUN tar -xzf gost_3.0.0-nightly.20240711_linux_arm64.tar.gz -C tmpfiles && mv tmpfiles/gost /bin/gost && chmod +x /bin/gost
+                - RUN wget -q https://github.com/go-gost/gost/releases/download/v3.0.0-nightly.'$$cap_gost_nightly'/gost_3.0.0-nightly.'$$cap_gost_nightly'_linux_amd64.tar.gz
+                - RUN tar -xzf gost_3.0.0-nightly.'$$cap_gost_nightly'_linux_amd64.tar.gz -C tmpfiles && mv tmpfiles/gost /bin/gost && chmod +x /bin/gost
                 -  # Remove temp files
-                - RUN rm -rf tmpfiles gost_3.0.0-nightly.20240711_linux_arm64.tar.gz
+                - RUN rm -rf tmpfiles gost_3.0.0-nightly.'$$cap_gost_nightly'_linux_amd64.tar.gz
```
githubsaturn, line 24: Did you test this and it worked? Technically this should not work as the quotation marks will be in the string once $$cap_gost_nightly is replaced.

We should remove the quotation marks.

amirsaam: yes it does work, because the urls for tar files have "_" after the variable, it will break w/o quotation marks.

githubsaturn:
Thanks for confirmation!

PR #1136, opened 2024-07-21
Mautic v5 Update - Always clear cache in Service Update Override

AdamDudley:
I've noticed when rebooting Caprover server or restarting app, mautic wants the cache cleared - adding to startup script.

Mautic will show a most annoying message

 "The site is currently offline due to encountering an error. If the problem persists, please contact the system administrator"

And nothing else. in Mautic forums, everybody suggests just to clear the cache.

This update just does that every time, as there really is no negatives, you've restarted the server anyway, what's another few seconds to get loaded up.

https://forum.mautic.org/t/the-site-is-currently-offline-due-to-encountering-an-error-if-the-problem-persists-please-contact-the-system-administrator/24199/1

- ☑️ I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
-☑️ I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).

### DIFF
--- a/public/v4/apps/mautic.yml
+++ b/public/v4/apps/mautic.yml
@@ -86,7 +86,7 @@ caproverOneClickApp:
             b. In "Environmental Variables", ensure these variables are set correctly:
                 - APACHE_SERVER_NAME = $$cap_appname.$$cap_root_domain
             c. Find the "Service Update Override" section.
-            d. Copy and paste this script:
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #1137, opened 2024-07-21
Formbricks - Added all SMTP environment vars, updated instructions and set example…

AdamDudley:
Formbricks - Added all SMTP environment vars, updated instructions and set example so user's can just click next and formbricks still works

apologies if this is a duplicate pull request, had my accounts confused - please consider this one the correct one.

- ☑️I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- ☑️ I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).


githubsaturn:
Thanks!

PR #1150, opened 2024-09-20
Fixed MAIL_DRIVER default option for Docmost app

larananas:
Hi,

I'm really sorry about the issue with my previous pull request. I didn't mean for that to happen, and this new PR fix the problem. I've corrected the default MAIL_DRIVER option for the Docmost app, which was causing the app to crash in a loop on startup without this environment variable set.

I tested everything from scratch, and it's all working fine now.

Apologies again for the trouble, and thanks for your understanding. 🙇‍♀️


githubsaturn:
Thanks so much for the fix!

PR #1156, opened 2024-10-11
{Plausible}change to smtp image from bytemark/smtp	to navanchauhan/smtp to include image with arm support 

abhranil26:
Also updated default version of plausible to the latest version.

I have tested this, it works!

- [ ] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- [ ] I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).

### DIFF
--- a/public/v4/apps/plausible.yml
+++ b/public/v4/apps/plausible.yml
@@ -2,7 +2,7 @@ captainVersion: 4
 
 services:
     $$cap_appname-mail:
-        image: bytemark/smtp
+        image: navanchauhan/smtp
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks for the PR, but this image is very random image and we won't include it in the official repository due to high risk. 

The official image is much safer - even though it may not have the capabilities of the other image.

abhranil26:
Alright!

PR #1157, opened 2024-10-12
[Plausible]Removed SMTP container uses Plausible's inbuilt Bamboo.mua for emails.

abhranil26:
Checked and it works!!
Emails are being sent.


githubsaturn:
thanks!

PR #1181, opened 2024-11-28
Adding: LinkAce

wojtishek:

- [ ] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)


githubsaturn:
Thanks!

VictorioBerra:
@wojtishek Can we not append `-simple` to LinkAce tags? I am unable to use LinkAce 2.x because of this. There is no `2.x-simple`. Is there a reason we require this?

VictorioBerra:
@wojtishek ARe we unable to use Beta? Should we create a new template for LinkAce-Beta?

wojtishek:
@VictorioBerra better solution should be to create Linkace-simple and Linkace-advanced templates IMHO (https://www.linkace.org/docs/v1/setup/setup-with-docker/). I'm not Linkace developer, so not sure what type of deployment is beta version.

wojtishek:
v2 is out, but changed image tags, I'll update one click app

@githubsaturn what will be best approach? leave current app, rename to v1 and create v2 app? depending upgrade guide there's one BC - `MAIL_DRIVER needs to be replaced with MAIL_MAILER, but the values stays the same. Example: MAIL_DRIVER=smtp becomes MAIL_MAILER=smtp`

githubsaturn:
I recommend just changing the current one to v2. No need to keep v1.

wojtishek:
#1199

PR #1183, opened 2024-12-10
Added LangflowAI

AdamDudley:

- [ ☑️ ] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [☑️  ] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [ ☑️ ] Icon is added as a png file to the logos directory.
- [☑️  ] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- [ ☑️ ] I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).


public/v4/apps/langflow.yml
```diff
@@ -0,0 +1,134 @@
...
+          defaultValue: $$cap_gen_random_hex(32)
+          description: Secret key used to encrypt sensitive data.
+          validRegex: /.{32,}/
+        - id: $$cap_langflow_port
+          label: LangFlow Port
+          defaultValue: 7860
+          description: Port on which the LangFlow server will run.
+          validRegex: /^\d+$/
+        - id: $$cap_langflow_database_url
+          label: PostgreSQL Database URL
```
githubsaturn, line 54: Let's include postgres as a service in this template. One click apps are meant to be truly one click as we should try to keep all the dependencies in the same template.

public/v4/apps/langflow.yml
```diff
@@ -0,0 +1,134 @@
...
+          label: Superuser Password
+          defaultValue: $$cap_gen_random_hex(16)
+          description: Password for the superuser. Edit this if needed.
+          validRegex: /.{1,}/
+        - id: $$cap_langflow_secret_key
+          label: Secret Key
+          defaultValue: $$cap_gen_random_hex(32)
+          description: Secret key used to encrypt sensitive data.
+          validRegex: /.{32,}/
+        - id: $$cap_langflow_port
```
githubsaturn, line 48: This doesn't really need to be a variable, let's fix it to 7860. Since it's not mapped and it's internal to the container, it's totally fine to keep it as a fixed value.

public/v4/apps/langflow.yml
```diff
@@ -0,0 +1,134 @@
...
+        environment:
+            LANGFLOW_SUPERUSER: $$cap_langflow_superuser
+            LANGFLOW_SUPERUSER_PASSWORD: $$cap_langflow_superuser_password
+            LANGFLOW_SECRET_KEY: $$cap_langflow_secret_key
+            LANGFLOW_DATABASE_URL: $$cap_langflow_database_url
+            LANGFLOW_SAVE_DB_IN_CONFIG_DIR: true
+            LANGFLOW_AUTO_LOGIN: $$cap_langflow_auto_login
+            LANGFLOW_BACKEND_ONLY: $$cap_langflow_backend_only
+            LANGFLOW_DEV: $$cap_langflow_dev
+            LANGFLOW_OPEN_BROWSER: $$cap_langflow_open_browser
```
githubsaturn, line 17: Since could be hardcoded to false. Since it's run in a container, it won't launch the browser anyways.

githubsaturn (commented):
Awesome app! Thanks!

public/v4/apps/langflow.yml
AdamDudley: yes good point. I'll update that in a little while

PR #1185, opened 2024-12-18
Update mysql.yml

ThcDelux3:
Upgrade MySQL from EOL version 5.7 to latest LTSR

MySQL 5.7 reached End of Life in October 2023 (see https://en.wikipedia.org/wiki/MySQL). Upgrade to the latest supported LTSR version to ensure continued security updates, compatibility, and stability.


githubsaturn:
Thanks!

PR #1188, opened 2024-12-31
remove emailengine as it is not open source anymore

lexoyo:
as [stated in this discussion](https://github.com/caprover/one-click-apps/pull/676#issuecomment-2566258985), emailengine is not open source anymore, only an outdated branch is

I believe it does not have its place in the community repo

- [-] I have tested the template using the method described in README.md thoroughly
- [-] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [-] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [-] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [-] Icon is added as a png file to the logos directory.


githubsaturn:
Thanks! 

Just for the reference, the one click apps do not need to be necessarily open source. But they at least need to provide a FREE functionality. In this case, it looks like the app doesn't have any FREE functionalities at all.

PR #1189, opened 2025-01-02
Fix docker hub url for valkey

rafael1mc:
Noticed that the docker hub url for valkey was leading to a 404 page, so I fixed it.

I looked for others, and this seems to be the only app not having either `/r` or `/_`.

<hr>


githubsaturn:
Thanks!

PR #1197, opened 2025-02-03
Fix n8n persistant dir

lexoyo:
This fixes the persistant directory, which is wrong initially

This is especially anoying because if you fix the path after having used n8n for a while it messes up the ecryption key of all your services (credentials, apis etc) which resets all your integrations


githubsaturn:
Thanks!!

PR #1206, opened 2025-03-10
[Draft] Update mautic.yml with embedded containers for worker and cron modes

whybtech:
Thanks for all the good apps available with one click! 

Actual version of Mautic template do not deploy crontab even if **MAUTIC_RUN_CRON_JOBS** is set to true (maybe deprecated),  you have to implement it manually.

Mautic 5 recommands to use 3 containers according the 3 modes : web, cron and worker
See example for docker-compose.yml : [https://github.com/mautic/docker-mautic/blob/mautic5/examples/basic/docker-compose.yml](https://github.com/mautic/docker-mautic/blob/mautic5/examples/basic/docker-compose.yml)

I tested after app is installed to create contact and segment with some contact data to match my filter and after waiting few seconds until `mautic:segments:update` task is executed, my contact is now automatically added to segment

### DIFF
--- a/public/v4/apps/mautic.yml
+++ b/public/v4/apps/mautic.yml
@@ -21,18 +21,56 @@ services:
             - $$cap_appname-config:/var/www/html/config
             - $$cap_appname-logs:/var/www/html/var/logs
             - $$cap_appname-media:/var/www/html/docroot/media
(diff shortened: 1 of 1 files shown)


githubsaturn:
Thanks!

PR #1209, opened 2025-03-23
Static Website One Click App using Alpine + Apache

amirsaam:


githubsaturn:
While I see the merit for this, it doesn't really qualify as a one click app installations.

One click apps are the apps that are installed and are ready to be used. Whereas this particular one is a just a shell apache container that serves files from a volume.

githubsaturn:
It's like simply deploying a new app with `nginx:latest` and have a folder mapped to `/usr/share/nginx/html`... Not really a one click app as you still need to manually copy your content to that volume.

PR #1213, opened 2025-04-08
Add Bugsink template

vanschelven:
I did the self-checks. I did not understand this one: running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`). It says "if necessary". Is it?

Some thoughts/doubts of my own before this gets merged:

* Is it possible to auto-enable https from the one-click installer? The template, as it stands, more or less depends on it.
* Should we add SMTP (and possibly other) variables to the template?
* There is a small window in time (1 or 2 seconds) where the mysql database is not quite ready and the web db already being accessible (but giving a 500). Is this something to improve on, and is the solution known?


githubsaturn:
> Is it possible to auto-enable https from the one-click installer? The template, as it stands, more or less depends on it.

No, but you can add a note in the instructions.end saying that you MUST enable HTTPS to use this app.

> Should we add SMTP (and possibly other) variables to the template?

Up to you. But general rule of thumb is to keep installation simple - as long as these variables can be added later and do not HAVE to be included on the first run

> There is a small window in time (1 or 2 seconds) where the mysql database is not quite ready and the web db already being accessible (but giving a 500). Is this something to improve on, and is the solution known?

MySQL is deployed prior to the bugsink (perhaps 10-20 seconds before), I doubt this is the cause. But generally speaking, yes this is fine. The reason is that you cannot gaurantee that MySQL can run 100% of the time, a simple server crash might also result in a MySQL restart.

public/v4/apps/bugsink.yaml
```diff
@@ -0,0 +1,79 @@
...
+          label: Database | Version
+          description: MySQL version. Check out their valid tags at https://hub.docker.com/_/mysql/tags
+          defaultValue: '8.4.4'
+          validRegex: /.{1,}/
+        - id: '$$cap_MYSQL_PASSWORD'
+          label: MySQL Password
+          defaultValue: $$cap_gen_random_hex(16)
+          description: The password for the MySQL root user.
+          validRegex: '/.{1,}/'
+        - id: '$$cap_SUPERUSER_PASSWORD'
```
githubsaturn, line 60: Recommend using 
```
defaultValue: $$cap_gen_random_hex(16)
```
for easier installation

githubsaturn: Same for MySQL password

public/v4/apps/bugsink.yaml
```diff
@@ -0,0 +1,79 @@
...
+            * Sentry-SDK compatible
+            * Scalable and reliable
+            Documentation: https://www.bugsink.com/docs/
+        end: |-
+            Bugsink instance successfully deployed. Visit https://$$cap_appname.$$cap_root_domain/ to access your Bugsink instance.
+            Log in with admin/SUPERUSER_PASSWORD.
+            Post-install Quickstart: https://www.bugsink.com/docs/quickstart/
+    displayName: Bugsink
+    isOfficial: true
+    description: A relatively short description, less than 200 characters.
```
githubsaturn, line 78: ??

vanschelven: sloppy! will fix

public/v4/apps/bugsink.yaml
vanschelven: > `defaultValue: $$cap_gen_random_hex(16)`

that's what I wrote, isn't it?

for the superuser password, I intentionally left it blank to avoid the user setting something without noticing it, and then not being able to log in. But if that's not the way to go on caprover, I'll gladly change.

githubsaturn:
I'll merge this as is, but since you're the author, I recommend changing this:
> Bugsink hard-ties the ideas of "behind a proxy" and "use https" together in the belief that here is no reason to go through the trouble of setting up a proxy if you're not using https."

One usecase is CapRover in internal set up. But certainly there are more use cases.

PR #1225, opened 2025-05-29
Update appwrite one-click to v17

JonahGoldsmith:
I updated the appwrite.yml file for my own use-case and it works perfect. Thought maybe the community would like it


githubsaturn (commented):
Thanks for the PR. Just a couple of questions to make sure things are not accidentally removed.

This is perhaps the most complex app that we have on the official repo at this point. Almost 1300 lines of yaml... impressive!

PR #1230, opened 2025-07-03
Added Collabora Online - Independent

manjujidagi:


githubsaturn:
@manjujidagi - yes, but the same could be achieved by entering an empty value for `cap_wopihost_domain` in the existing app. As far as I can see, they are otherwise identical?

Am I missing a difference?

githubsaturn:
If that's the case, then all you need to do is to change these:
https://github.com/caprover/one-click-apps/blob/eebe69c5372f66af8a671111dee6f0b34f7a9a5b/public/v4/apps/collabora-online.yml#L33-L37

* Change valid regex such that it accepts empty string
* Add to the description that "enter empty string if you want to use it without any nextcloud installation"

manjujidagi:
@githubsaturn Updated accordingly

PR #1233, opened 2025-07-14
update Docker image to official evolutionapi channel

jatinsne:
Switched Docker image source to the official evolutionapi channel on Docker Hub.
This ensures alignment with the maintained and verified version of the API image.

Previous source: atendai/evolution-api
New source: evoapicloud/evolution-api

- [☑️] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- [☑️] I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).

### DIFF
--- a/public/v4/apps/evolution-api.yml
+++ b/public/v4/apps/evolution-api.yml
@@ -204,7 +204,7 @@ caproverOneClickApp:
     variables:
         - id: $$cap_ant_version
           label: Evolution API Version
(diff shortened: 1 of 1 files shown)


githubsaturn:
Not sure if this is a more official version, their site is still referencing the other image: https://doc.evolution-api.com/v2/en/install/docker

PR #1236, opened 2025-07-29
Upgrade default version of Bugsink to 1.7.4

vanschelven:
direct cause: https://www.bugsink.com/blog/path-traversal-via-event-id-in-ingestion/


vanschelven:
I have not run through the mentioned tests at this time; but I happen to know the installation requirements have not changed from 1.4.2 -> 1.7.4.

If you want to be more conservative, 1.4.3 would be a good bet (which is 1.4.2, but patched)

githubsaturn:
Thanks!

githubsaturn:
As long as it works on a new install, we're good

PR #1257, opened 2025-11-08
Enable SSL for Hedgedoc configuration

IzzleNizzle:
I used the one-click-app to install this application in my caprover instance. I immediately enabled https and websockets as i prefer https and assumed the app used websockets. 

However i was unable to register a user out of the box as the app was intended to do. Upon investigating the dev tools in chrome, it was complaining about http request to /register being blocked. 

I was unable to register a user upon first installation until i updated the env variable to use ssl. 

`CMD_PROTOCOL_USESSL`

I'm slightly conflicted, as i suppose someone might not want to use https. If that is the case, may i suggest that we add this information to the end instructions property instead. This may save someone the time it takes to debug and find solution from documentation like i did.

### DIFF
--- a/public/v4/apps/hedgedoc.yml
+++ b/public/v4/apps/hedgedoc.yml
@@ -9,6 +9,7 @@ services:
             CMD_DOMAIN: $$cap_appname.$$cap_root_domain
             CMD_URL_ADDPORT: false
             PLUGIN_INSTALLER: $$cap_plugin_installer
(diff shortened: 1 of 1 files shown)


IzzleNizzle:
Oh i was just about to add an issue for this and i saw the instructions. My bad

### App bugs:
For issues regarding an app (post installation), please report them in the related app forum. CapRover is simply an installation tool. Post installation bugs/issues should be redirected to the developer of the app.

PR #1260, opened 2025-11-15
Updated Affine for newer versions

KrunalSM:
Updated Affine application configuration with new image and environment variables. Adjusted service dependencies and health checks.

### DIFF
--- a/public/v4/apps/affine.yml
+++ b/public/v4/apps/affine.yml
@@ -1,6 +1,6 @@
 captainVersion: 4
 services:
-    # Main Affine Application
+    #affine:
     $$cap_appname:
         image: ghcr.io/toeverything/affine:$$cap_affine_version
         # No command is needed, uses the image's default CMD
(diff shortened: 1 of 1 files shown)


public/v4/apps/affine.yml
```diff
@@ -1,143 +1,165 @@
...
+        command: ['sh', '-c', 'node ./scripts/self-host-predeploy.js']
+        depends_on:
+            $$cap_appname-redis:
+                condition: service_healthy
+            $$cap_appname-postgres:
+                condition: service_healthy
+        volumes:
+            - $$cap_appname-config:/root/.affine/config
+            - $$cap_appname-storage:/root/.affine/storage
+        restart: on-failure # Run as a one-off job
```
githubsaturn: This is ignored in CapRover. What would we the downside of running this migration potentially more than once? Is it smart enough to do nothing the second time it runs?

If so, you can change it to:
```
command: ["sh", "-c", "node ./scripts/self-host-predeploy.js && tail -f /dev/null"]
```

Also change the image name to `$$cap_appname-migration-delete-me-after-install`

githubsaturn: Also, is this really required even for a fresh install?

githubsaturn (commented):
Thanks for the PR!

githubsaturn:
Thanks!

PR #1286, opened 2026-04-05
Add Beszel one-click app

duylddev:
## Description

Add [Beszel](https://beszel.dev) - a lightweight server monitoring hub with historical data, Docker stats, and alerts. Built on PocketBase.

### Features
- Lightweight and fast monitoring dashboard
- Docker container stats tracking
- Historical data with customizable retention
- Alerts via email, webhook, and more
- Multi-server monitoring with agents

### Template Details
- **Image:** `henrygd/beszel`
- **Port:** 8090
- **Volume:** `/beszel_data` for persistent data

This template deploys the Beszel Hub only. The Beszel Agent should be installed separately on each monitored server.

### References
- Website: https://beszel.dev
- GitHub: https://github.com/henrygd/beszel
- Docker Hub: https://hub.docker.com/r/henrygd/beszel

## Self Check before Merge


VictorioBerra:
> Nice! We should probably move away from Netdata to this one as the built-in monitoring.
> 
> Do you happen to know if it supports email/telegram/slack notifications on CPU, RAM, Disk usage etc?

Is this still in the works? I think this is a great idea.

githubsaturn:
Not yet - but I have created an issue for visilbiy here: https://github.com/caprover/caprover/issues/2410


There are a few non trivial things that I should think of... e.g. what happens with legacy installs of Netdata... keeping them isn't easy as it creates a parallel dual system in the code base.... another option - which seems the right way - is to rethink the "monitoring" feature and make it just a one click app similar to regular apps... that way, multiple one click apps (monitoring apps) can live side by side each other without stepping on each others' toes

PR #1298, opened 2026-06-03
Updated GOST (with official image) + Removal of old GOST OCAs

amirsaam:
This update switches GOST one-click app to use official image of the software, also updates instruction to how to create the config file manually and use it.
+ It does remove old GOST one-click apps that used DockerFileLines to create the app (GOST v2 and v3)


public/v4/apps/gogost.yml
```diff
@@ -0,0 +1,56 @@
...
+                addr: ":$$cap_gost_api_port"
+                pathPrefix: "/api"
+                accesslog: true
+                auth:
+                    username: $$cap_gost_api_user
+                    password: $$cap_gost_api_password
+            ```
+            3. Go to CapRover dashboard and click **Save & Restart** to restart the app
+            4. Now you can start using **GOST API** in either of WebUI (https://ui.gost.run) or GUI (https://github.com/go-gost/gostctl/releases)
+            * Your GOST API would be accessible through http://$$cap_appname.$$cap_root_domain/$$cap_gost_api_path after step 3.
```
githubsaturn: `cap_gost_api_path` doesn't exist. Can you please update this instructions?

public/v4/apps/gogost.yml
```diff
@@ -0,0 +1,56 @@
...
+          description: The API port you configure in gost.yaml
+          defaultValue: 18080
+        - id: $$cap_gost_api_user
+          label: Username for GOST WebAPI
+          description: At least 4 characters
+          defaultValue: gostguru
+        - id: $$cap_gost_api_password
+          label: Password for GOST WebAPI
+          description: At least 8 characters, password should contain both uppercase and lowercase and at least one number and a special character (not at the start of password)
+          defaultValue: Ex@mple777
```
githubsaturn, line 39: Change to `$$cap_gen_random_hex(12)`

amirsaam: `$$cap_gen_random_hex(12)` doesn't match the `validRegex` so the field should be either empty or manually written

public/v4/apps/gogost.yml
amirsaam: removed by mistake, added api_path and api_alog back to be used in the post-deploy instruction

githubsaturn:
Thanks!

PR #1304, opened 2026-06-18
VitoDeploy one-click-app

RichardAnderson:
Adds the VitoDeploy (https://github.com/vitodeploy/vito) application to CapRover as a one-click-app.  

VitoDeploy allows the management of servers and sites, allowing you to automate the deployment of PHP, Node, Bun, Python and .NET applications from multiple source control providers.   We currently have 3.1k stars on GitHub and 10k+ downloads on DockerHub, and we are continuing to grow!  Vito deploy is actively maintained, and has been for several years. 

4.x tag selected as no breaking changes to be implemented within this stream that would cause issues with the deployment of VitoDeploy on CapRover, breaking changes will be released under the 5.x tag.   4.x is currently in stable beta, and a release of 4.0.0 will launch on the 1st July!


RichardAnderson:
As a follow up to this PR, v4 has released as stable, and the template continues to function.   Please do let us know if you have any concerns regarding adding VitoDeploy to the CapRover one-click-apps, we feel it would be a perfect addition.   

While they can potentially be seen as similar tooling - they do serve very different purposes, and many of us do already use CapRover to deploy our Vito instances right now, so having it as a one-click-app just makes that process even easier!

public/v4/apps/vitodeploy.yml
```diff
@@ -0,0 +1,68 @@
...
+caproverOneClickApp:
+    variables:
+        - id: '$$cap_vito_version'
+          label: Vito Version
+          defaultValue: '4.x'
+          description: Check out their Docker page for the valid tags https://hub.docker.com/r/vitodeploy/vito/tags (use a "-arm64" tag for ARM-based servers)
+          validRegex: '/^([^\s^\/])+$/'
+        - id: '$$cap_vito_app_key'
+          label: App Key
+          defaultValue: 'base64:$$cap_gen_random_hex(43)='
```
githubsaturn, line 26: Does this have to start with `base64:` and end with `=` ?

The documentation says just a [random 32 character](https://vitodeploy.com/docs/getting-started/installation#docker-compose:~:text=32%2Dcharacter%20string%20app%20key%20used%20for%20encryption%20within%20the%20app) is good

githubsaturn (commented):

public/v4/apps/vitodeploy.yml
RichardAnderson: Hi - it's a laravel app_key, so it does need to be base64, if you use the "Generate a random APP_KEY" button a little further up on the page, you will see it generates using the same format: `base64:CMw0viPqlf7HRSv3i1fCAlXFCd+PdkLZYjj51Mxhkto=` 

The docs could do with a bit of update in that regard.   But I can confirm that the key generation as per the yml here is correct.

githubsaturn:
Thanks!

PR #1323, opened 2026-08-13
Add Louez one-click app

loan-talvat:
## Louez

[Louez](https://github.com/Synapsr/Louez) is an open-source equipment rental platform (AGPL-3.0). Rent out cameras, tools, bikes, party gear or vehicles from a branded storefront with live availability, and manage products, reservations, pickups and returns, customers, PDF contracts and revenue statistics from one dashboard. The interface ships in 8 languages.

## What it installs

- `$$cap_appname` — `synapsr/louez` (storefront + dashboard + API, HTTP port 3000, creates its own database schema on first boot)
- `$$cap_appname-db` — `mysql:8.4`, not exposed as a web app
- `$$cap_appname-storage` — `minio/minio` for product images, not exposed as a web app

Passwords default to `$$cap_gen_random_hex(24)`, and the app's own auth secret is generated on first boot and persisted in its volume, so sessions survive redeploys.

## Testing

Tested before opening the PR. The same three-service stack was deployed and exercised end to end: services healthy, schema installed automatically, sign-up through the UI flow, product image uploaded through the app into the MinIO bucket and served back byte-identical. `npm run validate_apps` reports `Validated louez` and `npm run formatter` passes.

One detail worth flagging, because it bites every MinIO-based template: the bucket is created by the storage service's start command, and that command has to invoke the shell explicitly (`sh -c '…'`) — MinIO's entrypoint prepends `minio` to whatever arguments it receives, so anything else becomes an unknown sub-command and the container crash-loops. This matches the approach already used by `posthog.yml` in this repo.

-------------

- [ ] Please note that the official repository is for popular apps with 1k+ of stars or 10k+ downloads. If you'd like to add a less popular app, you can always [create your own 3rd party repo](https://github.com/caprover/one-click-apps#build-your-own-one-click-app-repository) and add your app there.


loan-talvat:
Thanks both — but I believe this finding is a false positive, and I'd rather show the evidence than just assert it.

The premise is correct: MinIO's entrypoint (`dockerscripts/docker-entrypoint.sh`) does prepend `minio` when the first argument isn't `minio`. I hit exactly that crash-loop while testing an earlier revision of this template that used a Compose-style `command` with the image's entrypoint intact.

What makes it a non-issue here is that CapRover does not pass `command` as *arguments* — it replaces the entrypoint. From `caprover/caprover`, `src/utils/DockerComposeToServiceOverride.ts`:

```ts
if (compose.command) {
    override.TaskTemplate = {
        ContainerSpec: {
            Command: Array.isArray(compose.command)
                ? compose.command
                : parseDockerCMD(compose.command),
        },
    }
}
```

In the Swarm API, `ContainerSpec.Command` is the ENTRYPOINT override (`ContainerSpec.Args` is the CMD equivalent) — it is what `docker service create --entrypoint` sets, documented as *"Overwrite the default ENTRYPOINT of the image"*.

Verified on a real Swarm node, using the exact ContainerSpec that `parseCommand()` produces for this template (full argv in `Command`, `Args` empty):

```
$ docker service inspect minio-caprover-shape \
    --format 'Command={{json .Spec.TaskTemplate.ContainerSpec.Command}} Args={{json .Spec.TaskTemplate.ContainerSpec.Args}}'
Command=["sh","-c","mkdir -p /data/louez && exec minio server /data"] Args=null

$ docker service ps minio-caprover-shape --format '{{.CurrentState}} {{.Error}}'
Running 17 seconds ago

$ docker exec <task> /bin/sh -c 'mc alias set local http://127.0.0.1:9000 louez … ; mc ls local/'
[0001-01-01 00:00:00 UTC]     0B louez/
```

The service starts and the bucket exists. `parseDockerCMD()` also splits the quoted string correctly — `["sh", "-c", "mkdir -p /data/louez && exec minio server /data"]` — because its regex keeps single-quoted sequences as one argument.

There is precedent in this repo too: [`public/v4/apps/posthog.yml`](https://github.com/caprover/one-click-apps/blob/master/public/v4/apps/posthog.yml) starts MinIO the same way (`command: sh -c 'mkdir -p /data/posthog && minio server …'`), and the README explicitly prefers `command` over `dockerfileLines`.

Happy to switch to a `dockerfileLines` wrapper (as `minio.yml` does) if maintainers prefer consistency with that template — but functionally the current form is correct and keeps the app to three containers instead of four.

githubsaturn:
Thanks for the pull request. Please follow the pull request template and ensure all boxes are checked.

loan-talvat:
Thanks for looking, and sorry for leaving the checklist blank — that was sloppy of me. I've filled it in on the PR description. Seven boxes are ticked; one isn't, and I'd rather explain it than quietly check it.

**On the popularity bar:** Louez is at **4,822 Docker pulls and 20 GitHub stars** as of today. That's about half the download threshold and nowhere near the star one, so the box stays unchecked.

So this is a request for an exception, not a claim of eligibility. What I can offer in support of it:

- **Honest trajectory.** Pulls are running at roughly 50/day, which puts the 10k mark about three months out at the current rate. I'd rather give you that number than call it explosive growth.
- **Actively developed, commercially backed.** Two major releases in the last six weeks (v2.0.0 on 23 July, v2.1.0 on 4 August). It's maintained by the company that also runs it as a hosted service, so the maintenance commitment in the last checkbox isn't hypothetical — there's a team whose product this is.
- **Now genuinely open source.** The repository was relicensed from Apache 2.0 + Commons Clause to **AGPL-3.0** yesterday, so it's OSI-approved rather than source-available. That was the main thing standing between Louez and the self-hosting ecosystem.
- **The template is finished and tested.** Full three-service stack deployed on Swarm and exercised end to end, pinned versions throughout, `validate_apps` and `formatter` both green.

One piece of it may be worth keeping regardless of what you decide here: the MinIO service needs its start command wrapped in an explicit `sh -c`, because MinIO's entrypoint prepends `minio` to whatever arguments it receives, so a bare command turns into an unknown sub-command and the container crash-loops. It's documented in the PR body and matches what `posthog.yml` already does.

If you'd prefer I come back once the download count clears 10k, that's entirely fair — say so and I'll close this and re-open then, no hard feelings. In the meantime I'll publish it as a third-party repo, as the checklist suggests.

githubsaturn:
Thank you for the response. As long as you're committed and maintaining this, I'm okay with merging for now.

PR #1303, opened 2026-06-13
Add LibreDB Studio one-click app

cevheri:
## Add LibreDB Studio

[LibreDB Studio](https://github.com/libredb/libredb-studio) is an open-source, web-based SQL IDE for the cloud era. It lets you query **PostgreSQL, MySQL, SQLite, Oracle, SQL Server, MongoDB and Redis** straight from the browser, with AI-powered query assistance, an ERD viewer, schema diff and a virtualized results grid.

### Template highlights
- Single service on container HTTP port **3000**.
- Pinned image `ghcr.io/libredb/libredb-studio` (default `0.9.14`, **no `:latest`**).
- **SQLite** persistence on a persistent volume (`$$cap_appname-data` → `/app/data`) — survives restarts and redeploys, no extra DB service.
- **Auto-generated** `JWT_SECRET` and admin/user passwords, **echoed in `instructions.end`** so users can log in right after install with minimal effort.
- Optional AI/LLM fields (gemini / openai / ollama / custom), blank by default.
- Advanced options (OIDC SSO, PostgreSQL storage backend) documented in `instructions.end` as post-install env vars.

### Validation
```
npm ci && npm run validate_apps   # Validated libredb-studio
npm run formatter                 # All matched files use Prettier code style!
```

---

### ☑️ Self Check before Merge

- [x] I have tested the template using the method described in README.md thoroughly
- [x] I have ensured that I put as much default values as possible (except passwords) to ensure minimum effort required for end users to get started.
- [x] I have ensured that I am not using the "latest" tag as this tag is dynamically changing and might break the one-click app. Use a fixed version.
- [x] I have made sure that instructions.start and instructions.end are clear and self-explanatory.
- [x] Icon is added as a png file to the logos directory.
- [x] I've executed the checks if necessary by running `npm ci && npm run validate_apps && npm run formatter` (If failling run the prettier: `npm run formatter-write`)
- [x] I will take responsibility addressing any issues that arises as a result of this PR (maintaining this app).

PR #1334, opened 2026-09-17
libredb-studio: bump default version to 0.16.0

yusuf-gundogdu:
The pinned default was still `0.9.59`; the current release is `0.16.0`. Bumps `defaultValue` and the matching example in the variable's description. Nothing else in the template changes.

Tested before submitting, on 0.16.0, by running the template's own image, environment and volume:

- `npm ci && npm run validate_apps` reports `Validated libredb-studio`, and `npm run formatter` reports `All matched files use Prettier code style!`
- The app comes up and `/login` returns 200.
- Login with the generated admin password returns 200 with the admin role; a wrong password returns 401.
- `STORAGE_PROVIDER=sqlite` takes effect: `/api/storage/config` reports `{"provider":"sqlite","serverMode":true}`, and `libredb-storage.db` is created at the path the template sets.
- A saved connection survives `docker restart` on the mounted volume, read back after the restart.

- [x] Please note that the official repository is for popular apps with 1k+ of stars or 10k+ downloads. If you'd like to add a less popular app, you can always [create your own 3rd party repo](https://github.com/caprover/one-click-apps#build-your-own-one-click-app-repository) and add your app there.

On the first box, for the record: 102,952 Docker Hub pulls and 787 GitHub stars at the time of writing, so it clears the downloads threshold rather than the stars one. The app itself was merged in #1303 and last bumped in #1315; this PR only moves the version.


yusuf-gundogdu:
One question while this is open, and it is about your review time rather than about this PR.

The image tag here is pinned, so a version bump is a one line change with no judgement in it. Two other stores that carry LibreDB Studio now watch the tag and open that bump themselves, so neither side writes or reviews it any more. One of their maintainers put it plainly: updates are handled automatically.

If you would like the same here, say the word and I will send whatever your setup needs, a renovate entry or the equivalent, as a separate small PR. If you would rather keep it manual, that is fine too and I will keep opening these by hand.

I am asking because this is the kind of change that does not really need a person on either end, and I would rather your attention went to the PRs that do.

githubsaturn:
we'll keep this manual as a future version might require a different env vars etc. When someone sends a PR we know that version works.

PR #1335, opened 2026-09-20
libredb-studio: bump default version to 0.16.1 and fix login over plain HTTP

yusuf-gundogdu:
Bumps the pinned default from 0.16.0 to 0.16.1, and fixes login on a fresh install.

### Login over plain HTTP

CapRover serves a new app at `http://<app>.<root-domain>` until the operator enables HTTPS from the dashboard, which our own post-install text already says. The app runs with `NODE_ENV=production` and marks its auth cookie `Secure` on any non-loopback host, so the browser discards the cookie and login bounces back to the sign-in page with the correct password and no error.

`AUTH_COOKIE_SECURE: 'false'` is now set, with a comment telling the operator to switch it to true once HTTPS is on. This is not a regression from the version bump. The behaviour was there before, and 0.15.0 was the first release that offered an override.

### Cleanup

The file carried an emoji and three em-dashes in its comments and post-install text. Those are gone, the wording is otherwise unchanged.

### Tested

Brought the app up with this file's defaults against `ghcr.io/libredb/libredb-studio:0.16.1`:

- container came up and stayed up
- `/api/storage/config` answered `{"provider":"sqlite","serverMode":true}`
- admin login returned `{"success":true,"role":"admin"}`
- the login cookie carries `HttpOnly; SameSite=lax` and no `Secure` flag, which is what makes login work before HTTPS is enabled
- the data volume at `/app/data` was written

`ghcr.io/libredb/libredb-studio:0.16.1` is the current release, published 19 September, and publishes linux/amd64 and linux/arm64.

### Self Check before Merge

- [x] Please note that the official repository is for popular apps with 1k+ of stars or 10k+ downloads. If you'd like to add a less popular app, you can always [create your own 3rd party repo](https://github.com/caprover/one-click-apps#build-your-own-one-click-app-repository) and add your app there.

On the first box, for the record: 110872 Docker Hub pulls and 821 GitHub stars at the time of writing, so it clears the downloads threshold rather than the stars one. The app was merged in #1303 and last bumped in #1334. On the last two boxes: the logo is already in the logos directory from #1303 and unchanged here, and `npm ci && npm run validate_apps && npm run formatter` was run on this branch, reporting `Validated libredb-studio` and `All matched files use Prettier code style!`, both exiting 0.

## Summary by CodeRabbit

- **Updates**
  - Updated the default LibreDB Studio version to 0.16.1.
  - Added a setting to control whether authentication cookies use the Secure attribute, remaining disabled until HTTPS is enabled.
  - Updated the example configuration and app description to reflect the new version.
  - Simplified installation completion messaging by removing decorative icons.


yusuf-gundogdu:
Updated to 0.16.2, which is the current release.

0.16.2 closes a cross-account authorization defect (GHSA-3wh2-8x78-jfw4, scored 8.8): a connection id supplied in a request could select another session's already-authenticated database provider. It matters on any instance more than one person signs in to, so this is worth taking ahead of an ordinary version bump.

I also pulled the template back in line with the source it is generated from upstream. The engine list now names sixteen engines rather than seven, the AI variable descriptions match what the app actually does, and the app description is 197 characters so it fits the catalog limit. AUTH_COOKIE_SECURE stays, since CapRover serves over http until HTTPS is enabled and the login loops without it.

Repo checks run locally: validate_apps says "Validated libredb-studio" and the Prettier check passes.

